PR #61, opened 2025-09-18
First Draft of Frightened and Antagonize Automations

mustangsami:


src/effects/frightened.ts
```diff
@@ -0,0 +1,119 @@
...
+    } else if (currentValue > 1) {
+        await condition.decrease();
+    } else {
+        logd(`Frightened value of ${currentValue} not expected.`);
+    }
+}
+
+async function createFrightenedRemovalConfirmationChatMessage(actor: ActorPF2e, antagonizeEffect: EffectPF2e, frightenedCondition: ConditionPF2e) {
+
+    const antagonizerId = antagonizeEffect?.flags?.samioli?.antagonizer as string | undefined;
```
olilan1, line 45: I think the question marks are not needed after `antagonizeEffect` or `flags`

src/effects/frightened.ts
```diff
@@ -0,0 +1,119 @@
...
+    const antagonizer = antagonizerId ? game.actors.get(antagonizerId) : undefined;
+    const antagonizerName = antagonizer?.prototypeToken?.name ?? "Unknown";
```
olilan1, line 47: Why do we not use the actor name if the token name is not available, like you do in your other code?

If you added the display name function per my other suggestion, you could use it here.

src/effects/frightened.ts
```diff
@@ -0,0 +1,119 @@
...
+
+    const content = `
+    <p>${actor.name} is currently Antagonized by ${antagonizerName}.</p>
```
olilan1, line 50: maybe remove "currently"

src/effects/frightened.ts
```diff
@@ -0,0 +1,119 @@
...
+    <p>Has ${actor.name} taken a hostile action against them, or can they no longer observer or sense them for at least one round?</p>
```
olilan1, line 51: Perhaps "or have they been unable to observe or sense ${actor.name} for at least one round?" would be more grammatical

src/effects/frightened.ts
```diff
@@ -0,0 +1,119 @@
...
+    <div style="display: flex; justify-content: center; gap: 10px; margin-top: 10px;">
+        <button type="button" data-action="remove-frightened-and-antagonize" 
+        data-frightened-condition-id="${frightenedCondition.id}" 
+        data-actor-id="${actor.id}" 
+        data-antagonized-effect-id="${antagonizeEffect.id}">
+            Remove Frightened and Antagonize
```
olilan1, line 57: As this is meant to be pressed in response to the question above, perhaps the text of the button should reflect that. In other words, perhaps the label should be "Yes, remove Frightened and Antagonize" (or just "Yes, remove Frightened")

src/effects/frightened.ts
```diff
@@ -0,0 +1,119 @@
...
+        speaker: ChatMessage.getSpeaker({ actor: actor }),
+        flags: {
+            samioli: {
+                buttonSlug: `remove-frightened-and-antagonize-button`
+            }
+        }
+    });
+}
+
+export function checkIfFrightenedAndAntagonizeButtonMessage(chatMessagePF2e: ChatMessagePF2e, html: JQuery<HTMLElement>) {
```
olilan1, line 76: Perhaps "addClickHandlerToFrightenedAndAntagonizeButtonIfNeeded"

src/effects/frightened.ts
```diff
@@ -0,0 +1,119 @@
...
+    const buttonSlug = chatMessagePF2e.flags?.samioli?.buttonSlug;
```
olilan1, line 77: no need for question mark after flags

src/effects/frightened.ts
```diff
@@ -0,0 +1,119 @@
...
+                const { frightenedConditionId, actorId, antagonizedEffectId } = button.dataset;
+                if (frightenedConditionId && actorId && antagonizedEffectId){
+                    handleRemoveFrightenedAndAntagonize(frightenedConditionId, actorId, antagonizedEffectId);
+                }
+            });
+        }
+    }
+}
+
+async function handleRemoveFrightenedAndAntagonize(frightenedConditionId: string, actorId: string, antagonizedEffectId: string) {
```
olilan1, line 94: just "removeFrightenedAndAntagonize" might be ok as a name?

src/effects/frightened.ts
```diff
@@ -0,0 +1,119 @@
...
+       
+    const token = canvas.tokens.placeables.find(t => t.actor?.id === actorId);
```
olilan1, line 96: Is there a reason you're using the token to get this, rather than doing `game.actors.get` like you do elsewhere?

src/effects/frightened.ts
```diff
@@ -0,0 +1,119 @@
...
+    const actor = token?.actor;
+
+    if (!actor || !actor.isOwner) {
+        ui.notifications.warn("You do not have permission to remove this effect.");
+        return;
+    }
+
```
olilan1, line 103: nit: double new line

src/effects/frightened.ts
```diff
@@ -0,0 +1,119 @@
...
+
+    const condition = actor.items.find(item => item.id === frightenedConditionId) as ConditionPF2e;
+    if (!condition) {
+        logd(`Could not find the frightened condition on the actor.`);
```
olilan1, line 107: I guess this case shouldn't happen, but if it did happen, shouldn't we still remove the antagonized effect if it's present?

So maybe:

``` TS
if (condition) {
    await condition.delete();
} else {
    logd(`Could not find the frightened condition on the actor.`);
}
```

src/actions/antagonize.ts
```diff
@@ -0,0 +1,141 @@
+import { ActorPF2e, ChatMessagePF2e, CheckContextChatFlag, ConditionPF2e, EffectPF2e } from "foundry-pf2e"
+import { getOwnersFromActor } from "../utils.ts";
+
+export async function checkForAntagonizeFeat(chatMessage: ChatMessagePF2e) {
+    const context = chatMessage?.flags?.pf2e?.context as CheckContextChatFlag;
```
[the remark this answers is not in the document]
mustangsami: I swear it was complaining at me before.

src/effects/frightened.ts
mustangsami: Yes, there was some weirdness where getting the actor wasn't bringing the right actor up for me. It was rather odd. This was the only way I could get it to work.

src/effects/frightened.ts
```diff
@@ -1,66 +1,77 @@
...
 
-async function decrementFrightenedCondition(condition: ConditionPF2e<ActorPF2e>) {
-    
-    const currentValue = condition.value;
-    if (currentValue === null) {
-        return;
+    if (frightenedValue === 1 && hasAntagonizedEffects) {
+        for (const effect of antagonizedEffects) {
+            await createFrightenedAndAntagonizeRemovalConfirmationChatMessage(token, effect, frightenedCondition);
+        }
```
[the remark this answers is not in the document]
mustangsami: I was deliriously tired when I wrote it so no wonder it read wierdly. Surprised it even worked!

src/utils.ts
```diff
@@ -142,14 +142,33 @@ export async function deleteLightFromTemplate(template: MeasuredTemplateDocument
...
+    });
+}
+
+export function returnStringOfNamesFromArray(names: string[]): string {
+  if (names.length === 0) {
+    return "";
+  }
+
+  if (names.length === 1) {
+    return `${names[0]}`;
```
olilan1, line 167: This could be just `return names[0];`

src/actions/antagonize.ts
```diff
@@ -123,9 +121,11 @@ export function warnIfDeletedItemIsFrightenedWhileAntagonized(item: ItemPF2e) {
 
     if (!hasAntagonizeEffect) return;
 
-    const actorName = getDisplayNameFromActor(actor);
+    const actorName = actor.prototypeToken.name;
```
olilan1, line 124: The prototype token isn't necessarily what we want here, is it? If the condition belongs to a token for a monster that has been renamed, then we want that token's name, not the prototype.

I think this can be `(actor.token ?? actor.prototypeToken).name`, which means take actor.token if possible, otherwise the prototypeToken, and then get the name.

src/effects/frightened.ts
```diff
@@ -132,20 +126,17 @@ async function removeFrightenedAndAntagonize(frightenedConditionId: string, toke
 
 async function sendFrightenedReducedMessage(actor: ActorPF2e, frightenedValue: number) {
     const recipients = getOwnersFromActor(actor).map(user => user.id);
-    const actorName = getDisplayNameFromActor(actor);
-    let content = ``;
+    const actorName = actor.prototypeToken.name;
```
olilan1, line 129: Again, prototype token probably isn't what we want here.

Maybe we should have a getDisplayName function, which can do `(actor.token ?? actor.prototypeToken).name`

olilan1 (changes requested):

PR #67, opened 2025-10-02
Damage-helper-poc

mustangsami:


src/damagehelper.ts
```diff
@@ -239,6 +94,186 @@ function showDamageHelperDialog() {
...
+
+    createDamageRoll(formData);
+    dialog.close();
+}
+
+function createLocalizedContexts(options: Record<string, string>): 
+    { value: string, label: string }[] {
+
+    const localizedContext = Object.entries(options).map(([key, labelKey]) => {
+        const value = key;
```
olilan1: this const seems unnecessary - just use `key` instead of `value` below?

src/damagehelper.ts
```diff
@@ -239,6 +94,186 @@ function showDamageHelperDialog() {
...
+
+    const value = formData.damageAmount as string;
+    const damageType = formData.damageType as string;
+    const material = formData.preciousMaterial as string;
+    const isPersistent = formData.persistent === true;
+    const damageTraits: DamageTrait[] = [];
+    const damageModifiers: string[] =[];
+
+    for (const key of Object.keys(damageTagConfig) as Array<keyof typeof damageTagConfig>) {
+        if (formData[key] === true) {
```
olilan1: No need to write `=== true` (ever)

src/damagehelper.ts
```diff
@@ -239,6 +94,186 @@ function showDamageHelperDialog() {
...
+            switch (group) {
+                case 'modifier':
+                    damageModifiers.push(tagValue);
+                    break;
+                case 'trait':
+                    damageTraits.push({ key: key, ...damageTagConfig[key] });
+                    break;
+            }
+        }
+    }
```
olilan1: This can be written a little more concisely by putting the value in the for loop as well as the key. Then you won't need to do `damageTagConfig[key]`.

Like this:

``` TS
    for (const [key, config] of Object.entries(damageTagConfig)) {
        if (formData[key]) {
            switch (config.group) {
                case 'modifier':
                    damageModifiers.push(config.value);
                    break;
                case 'trait':
                    damageTraits.push({ key, ...config });
                    break;
            }
        }
    }
```

src/damagehelper.ts
```diff
@@ -239,6 +94,186 @@ function showDamageHelperDialog() {
...
+    for (const key of Object.keys(damageTagConfig) as Array<keyof typeof damageTagConfig>) {
+        if (formData[key] === true) {
+            const { group, value: tagValue } = damageTagConfig[key];
+            switch (group) {
+                case 'modifier':
+                    damageModifiers.push(tagValue);
+                    break;
+                case 'trait':
+                    damageTraits.push({ key: key, ...damageTagConfig[key] });
+                    break;
```
olilan1: I know you handle persistent later, and it works because it's the only "type" value, but it would be more consistent to handle that here as well, by adding a `case 'type'` and pushing into `damageAndMaterialTypes`

src/damagehelper.ts
```diff
@@ -23,98 +36,37 @@ export function addDamageHelperButtonToChatUI(cssMappings: Record<string, HTMLEl
...
+    logd("Foundry v12 is not yet supported for the damage helper button.");
+}
 
-    const damageTypesOptions = Object.entries(damageTypes).map(([key, _labelKey]) => {
-        const localizedDamageLabel = localizedDamageLabels[key];
-        return `<option value="${key}">${localizedDamageLabel}</option>`;
-    }).join('');
+async function showDamageHelperDialog() {
 
+    const damageTypes: Record<string, string> = CONFIG.PF2E.damageTypes;
```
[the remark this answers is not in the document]
mustangsami: I attempted to do something with this but it took a LOT of rework of the rest of the code. Might be more suitable for you to walk me through this.

src/damagehelper.ts
```diff
@@ -239,6 +94,186 @@ function showDamageHelperDialog() {
...
+}
+
+function getFlavorHtml(damageType: string, damageTraits: { key: string; label: string; 
+    group: string; value: string; hasTag: boolean; }[] , material: string): string {
+
+    const hasTagTraits = damageTraits.filter(trait => trait.hasTag);
+    const noTagTraits = damageTraits.filter(trait => !trait.hasTag);
+    const headerStart = `
+        <h4 class="action">
+        <strong>Custom Damage Roll:
```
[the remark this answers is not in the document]
olilan1: The way I'm thinking about it is that this message is visible to all the players, not just the GM. I'm not sure the players need to know that I made a custom roll using this button rather than by typing "/r 1d6[fire]", and the technicality that everything is a "damage roll" won't mean anything to the players.

I get that you want something there, but I do think just "Damage:" would be fine, and it would make it less likely that the text will wrap onto the next line, which doesn't look so good.

src/damagehelper.ts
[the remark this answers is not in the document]
olilan1: OK, yeah we could do this as a follow-up

PR #74, opened 2025-10-29
Take Dread Effects into consideration when reducing Frightened

mustangsami:


src/effects/frightened.ts
```diff
@@ -16,25 +16,44 @@ export async function handleFrightenedAtTurnEnd(combatant: CombatantPF2e) {
     const frightenedValue = frightenedCondition.value;
     if (frightenedValue === null) return;
 
+    const despairEffects = getActorDespairEffects(actor);
+    const hasDespairEffects = despairEffects.length > 0;
```
olilan1: You don't appear to be using `despairEffects` anywhere, so you could instead have the function be `hasDespairEffects(actor)`, and in the function, use `some(blah)` rather than `filter(blah)`

PR #75, opened 2025-11-02
Unstable Check v1

mustangsami:


src/effects/unstablecheck.ts
```diff
@@ -1,14 +1,24 @@
-async function applyUnstableEffect(actor) {
+import { ActorPF2e, ChatMessagePF2e } from "foundry-pf2e";
+
+async function applyUnstableEffect(actor: ActorPF2e) {
     const unstableEffectItemId = "olpkQDGDzmYZCvQH";
     const compendiumPack = game.packs.get("pf2e.feat-effects");
+    if (!compendiumPack) return;
```
olilan1: Don't need this line if you put a question mark after `compendiumPack` on the next line

src/unstablehomebrew.ts
```diff
@@ -0,0 +1,164 @@
+import { ActorPF2e, ChatMessagePF2e, EffectSource } from "foundry-pf2e";
+import { ImageFilePath } from "foundry-pf2e/foundry/common/constants.mjs";
+import { addOrUpdateEffectOnActor, performFlatCheck } from "./utils.ts";
+import { createChatMessageWithButton } from "./chatbuttonhelper.ts";
+
+export function checkIfUnstableActionAndHandle(chatMessage: ChatMessagePF2e, html: JQuery<HTMLElement>) {
```
olilan1: sorry, I still don't like function names that being with "check".

Perhaps "replaceUnstableCheckWithStrainCheck" ?

src/unstablehomebrew.ts
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    let dc = 7;
+    const hasStrainEffect = actor.items.some(item => 
+        item.type === "effect" && item.system.slug === "samioli-strain"
+    );
+
+    if (hasStrainEffect) {
+        const strainEffect = actor.items.find(item => item.type === "effect" 
+        && item.system.slug === "samioli-strain");
```
olilan1: You're finding this effect twice here. Instead of the `some` call, you could just do the 'find', and check if it found anything.

src/unstablehomebrew.ts
```diff
@@ -0,0 +1,164 @@
...
+        
+        if (strainEffect) {
+            const strainValue = strainEffect.system.level?.value;
+            if (strainValue) {
+                dc += (strainValue * 5);
+            }
+        }
```
olilan1: All of these lines could instead be:

```TS
    dc += strainEffect?.system.level?.value ?? 0;
```

Also if you make the other suggested change to just use `find` once, you won't need an if statement at all, you can do this:

```TS
    const strainEffect = actor.items.find(item => 
        item.type === "effect" && item.system.slug === "samioli-strain"
    );
    dc += strainEffect?.system.level?.value ?? 0;
```

src/unstablehomebrew.ts
```diff
@@ -0,0 +1,164 @@
...
+
+    const rollOptions = chatMessage.flags.pf2e?.context?.options;
+    if (!rollOptions || !rollOptions.includes("samioli-unstable-check")) return;
+
+    const actor = chatMessage.actor;
+    if (!actor) return;
+
+    if (chatMessage.flags?.pf2e?.context?.outcome === "failure" 
+    || chatMessage.flags?.pf2e?.context?.outcome === "criticalFailure" ) {
+        // Unstable effect logic is handled by unstable check automation in unstablecheck.ts
```
olilan1: On first reading of this comment it wasn't obvious to me what it meant (I wasn't sure what "unstable effect logic is handled" meant).

Perhaps "Note that unstable effect is applied by checkForUnstableCheck function in unstablecheck.ts"

src/unstablehomebrew.ts
```diff
@@ -0,0 +1,164 @@
...
+        const strainEffect = actor.items.find(item => item.type === "effect" 
+        && item.system.slug === "samioli-strain");
```
olilan1: nit: indent 4 spaces

src/unstablehomebrew.ts
```diff
@@ -0,0 +1,164 @@
...
+        if (chatMessage.flags?.pf2e?.context?.outcome === "criticalFailure") {
+            createFireDamageChatMessage(actor);
+        }
+        if (!strainEffect) return;
+        strainEffect.delete();
+    } else {
+        const strainEffect = actor.items.find(item => item.type === "effect" 
+        && item.system.slug === "samioli-strain");
+        if (strainEffect) {
+            const currentLevel = strainEffect.system.level?.value || 1;
```
olilan1: Using `||` looks a bit unusual to me. Did you do this just to cover the case where value is null? If so, using `??` might be better.

I think the difference is that `value || 1` will be 1 if value is 0. Not sure if that's what you intended.

src/unstablehomebrew.ts
```diff
@@ -0,0 +1,164 @@
...
+        flavor: `On a critical failure for an unstable check, you take fire damage equal to half your level.`,
+        speaker: ChatMessage.getSpeaker({ actor: actor }),
+        flags: {
+            pf2e: {
+                context: {
+                    type: "damage-roll",
+                    target: {
+                        actor: actor,
+                        token: actor.token
+                    }
```
olilan1: Is this bit meant to make the damage roll have the inventor appear as the target using the Toolbelt thing? Because it didn't seem toi do that for me when I tried it.

olilan1 (approved):

PR #76, opened 2025-11-03
Fix to Dazzling Display to have the GM Client apply the Immunity Effects

mustangsami:


src/hooks.ts
```diff
@@ -210,11 +211,18 @@ function handleChatMessagePostRoll(message: ChatMessagePF2e) {
                     .ifMessagePosterAndActorOwner()
                     .run();
             break;
+        case "custom":
+            hook(runDazzlingDisplayAutomationAsGM, message)
+                    .ifGM()
+                    .run();
+            break;
```
olilan1: It might be better to give this type a more specific name. Otherwise if sounds like we'll end up running the dazzling-display hook for any custom message we create in future.

As you're putting the type name in a field, it could be "custom-dazzling-display" and we could later have other custom types.

(I wonder if it might be better to separate out these message types so we don't mix the actual types like "action" with the custom types, but you don't have to make that change here)

PR #77, opened 2025-11-03
Boost-Eidolon

mustangsami:


src/spells/boosteidolon.ts
```diff
@@ -0,0 +1,225 @@
...
+export async function runBoostEidolonAutomation(chatMessage: ChatMessagePF2e) {
+
+    // Check if the spell is Boost Eidolon
+    if (!chatMessage.flags?.pf2e?.origin?.rollOptions?.includes("origin:item:slug:boost-eidolon")) return;
+
+    // Don't add button if summoner has no focus points remaining
+    const summonerActor = chatMessage.actor;
+    if (!summonerActor) return;
+    if (summonerActor.type !== "character") return;
+    const summonerCharacterActor = summonerActor as CharacterPF2e;
```
olilan1: Could add a type guard function instead of having these three lines

src/spells/boosteidolon.ts
```diff
@@ -0,0 +1,225 @@
...
+}
+
+async function reduceFocusPoints(summonerActor: CharacterPF2e) {
+    const currentFocusPoints = summonerActor.system.resources.focus.value;
+    if (currentFocusPoints > 0) {
+        await summonerActor.update({"system.resources.focus.value": currentFocusPoints - 1});
+    }
+}
+
+function getSpellTraditionByEidolonType(eidolonActor: ActorPF2e): string | null {
```
olilan1: By writing the return type here as `string | null`, you've widened it and made it less useful.

The return type should actually be "arcane" | "divine" | "occult" | "primal" | null.

You could define a type:

```TS
type Tradition = "arcane" | "divine" | "occult" | "primal";
```

and that would be your return type. This will be useful when you then pass it in to getSkillCheckByTradition.

src/spells/boosteidolon.ts
```diff
@@ -0,0 +1,225 @@
...
+    // Apply Boost Eidolon effect to the eidolon
+    await createBoostEidolonEffectOnActor(eidolonActor);
+
+    // Calculate standard DC based on summoner level
+    const standardDCByLevel = getLevelBasedDC(summonerActor.level);
+    if (!eidolonActor) return;
+    const tradition = getSpellTraditionByEidolonType(eidolonActor);
+    if (!tradition) return;
+    const skillCheckRequired = getSkillCheckByTradition(tradition);
+    if (!skillCheckRequired) return;
```
olilan1: This null check shouldn't be needed - see other comments

src/hooks.ts
```diff
@@ -203,7 +204,12 @@ function handleChatMessagePostRoll(message: ChatMessagePF2e) {
                     .ifMessagePoster()
                     .run();
             break;
-        case "spell": 
+        case "spell":
```
olilan1: You've changed the behaviour for the "spell" case here, and possibly broken the sustain spells feature.

Before, when it said this:
```TS
case "spell":
case "spell-cast":
   hook(...)
   break;
```
that meant that the hook would run for either "spell" or "spell-cast" (as execution falls through from one case to the next if there is no "break")

Now with your change, it will only run for "spell-cast".

src/spells/boosteidolon.ts
```diff
@@ -0,0 +1,230 @@
+import { ActorPF2e, CharacterPF2e, ChatMessagePF2e, EffectPF2e } from "foundry-pf2e";
+import { getLevelBasedDC, isCharacter, logd } from "../utils.ts";
+import { createChatMessageWithButton } from "../chatbuttonhelper.ts";
+
+type Tradition = "arcane" | "divine" | "occult" | "primal";
+type MagicSkillCheck = "arcana" | "religion" | "occultism" | "nature";
```
olilan1: Perhaps just `MagicSkill`, as these are skill names?

src/chatbuttonhelper.ts
```diff
@@ -34,6 +34,7 @@ type MessageSpec = {
     actor: ActorPF2e,
     content: string,
     button_label: string,
+    flags?: Record<string, unknown>,
```
olilan1: Ah, you added this yourself! This is nice actually, as it allows objects to be included here rather than just strings in the `params` field.

Not fot this PR, but we could make it easier to use by adding the flags to the parameters of the function that gets called here. I can take a look at this after this PR gets submitted.

PR #78, opened 2025-11-08
Redistribute Potential Automation

mustangsami:


src/spells/redistributepotential.ts
```diff
@@ -0,0 +1,289 @@
...
+    [key: string]: JSONValue | undefined; 
+};
+
+type RedistributePotentialTypes = "steal-heat" | "concentrate-heat";
+
+export async function startRedistributePotential(token: TokenPF2e, message: ChatMessagePF2e) {
+
+    // Check if spell is amped or not
+    if (!message.flags.pf2e?.context?.options) return;
+    const isAmped = message.flags.pf2e?.context?.options.includes("amp-spell");
```
olilan1: Don't need the question marks here as you've already checked that `message.flags.pf2e.context.options` exists.

src/spells/redistributepotential.ts
```diff
@@ -0,0 +1,289 @@
...
+    const damageType = redistributePotentialType === "steal-heat" ? "cold" : "fire";
+
+    const DamageRoll = CONFIG.Dice.rolls.find((r) => r.name === "DamageRoll");
+    if (!DamageRoll) return null;
+
+    const roll = new DamageRoll(`${value}[${damageType}]`);
+    return roll;
+}
+
+function getRedistributePotentialItemToken(token: TokenPF2e): ItemPF2e | null {
```
olilan1: It's a bit confusing that this function is named getBlahToken when it doesn't return a token. Just `getRedistributePotentialItem` would be ok?

Also, rather than having to do `as SpellPF2e` after calling this function at line 58, it would be cleaner to have this function return SpellPF2e by doing the `as` here.

src/spells/redistributepotential.ts
```diff
@@ -0,0 +1,289 @@
...
+    if (!concentrateHeatTemplateData) return;
+    const concentrateHeatTemplate = await createTemplate(concentrateHeatTemplateData);
+    const concentrateHeatTargetTokens = await getTemplateTokens(concentrateHeatTemplate);
+
+    // Clean up templates
+    stealHeatTemplate.delete();
+    concentrateHeatTemplate.delete();
+
+    // Capture additional context for the spell damage message
+    const redistributePotentialItem = getRedistributePotentialItemToken(token) as SpellPF2e;
```
olilan1: Do we need to go back to the token to find the right spell? I'd have thought that the spell should be retrievable from the chat message.

Is it just `message.item`?

src/spells/redistributepotential.ts
```diff
@@ -0,0 +1,289 @@
...
+        switch (spellRank) {
+            case "5": return "4d4";
+            case "6": return "5d4";
+            case "7": return "6d4";
+            case "8": return "7d4";
+            case "9": return "8d4";
+            case "10": return "9d4";
+        }
+    }
+}
```
olilan1: This also feels unnecessary - it should be possible to get this from the spell.

Can you use `spell.system.damage`?

src/spells/redistributepotential.ts
```diff
@@ -0,0 +1,289 @@
...
+        <hr>
+        <div class="tags modifiers"><span class="tag tag_transparent" data-visibility="gm">${damageFormula} Fire</span></div>
+        <ul class="notes">
+            <li class="roll-note"><strong>Failure Effect</strong> A creature that fails its save also becomes <a class="content-link" draggable="true" data-link="" data-uuid="Compendium.pf2e.conditionitems.Item.MIRkyAjyBeXivMa7" data-id="MIRkyAjyBeXivMa7" data-type="Item" data-pack="pf2e.conditionitems" data-tooltip="Item" data-tooltip-text="Condition Item"><i class="fa-solid fa-face-zany" inert=""></i> Enfeebled 1</a>from heat stroke until the start of your next turn.</li>
+        </ul>
+    `;
```
olilan1: This function is very long, and having these big strings here is a bit awkward.

Also, from a code organisation perspective it feels like they should be part of the `sendRolledRollToChat` function rather than this one.

Suggest moving them out. Possibly would be cleaner to have a separate method that takes `redistributePotentialType`, `isAmpedText`, and `damageFormula` and returns the flavour text.

src/spells/redistributepotential.ts
```diff
@@ -0,0 +1,289 @@
...
+
+    // Roll Steal Heat damage but don't send message yet
+    const stealHeatDamageRoll = rollDamage(damageFormula, "steal-heat");
```
olilan1: Does this line actually roll the damage, or does that happen when you call `roll.evaluate()`?

Perhaps the function should be called `createDamageRoll` or `buildDamageRoll` instead

src/spells/redistributepotential.ts
```diff
@@ -0,0 +1,289 @@
...
+    replaceTargets(targetIds);
+
+    const spellDamage = await spellItem.getDamage();
+    if (!spellDamage) return;
+
+    await roll.toMessage(
+        {
+            speaker: ChatMessage.getSpeaker(),
+            flavor: flavorText,
+            content: `hello`,
```
olilan1: not needed?

src/spells/redistributepotential.ts
```diff
@@ -0,0 +1,289 @@
...
+                    origin: originData,
+                },
+            },
+        },
+        { create: true }
+    );
+
+}
+
+async function selectLocation(redistributePotentialType: RedistributePotentialTypes, locationObject: Point | TokenPF2e, isAmped: boolean): Promise<Point | false> {
```
olilan1: No need to say locationObject: Point | TokenPF2e.

Logically, this is just a Point.

The reason it works when you pass in a token, is that a token has `x` and `y` properties, so a token can act as a Point. (TypeScript lets you do that, because that's how JavaScript works)

Also if it's just a point, maybe better to just call it `location`.

src/spells/redistributepotential.ts
```diff
@@ -0,0 +1,289 @@
...
+    width: number;
+    distance: number;
+    direction: number;
+    fillColor: `#${string}`;
+    borderColor: `#${string}`;
+    flags?: { [x: string]: { [x: string]: JSONValue } };
+    [key: string]: JSONValue | undefined; 
+};
+
+type RedistributePotentialTypes = "steal-heat" | "concentrate-heat";
```
olilan1: Type names should be singlular

src/spells/redistributepotential.ts
```diff
@@ -0,0 +1,289 @@
...
+
+    if (canvas.scene === undefined || canvas.scene === null) return null;
+    const offset = isAmped ? 0 : canvas.scene?.grid.size / 2;
+
+    const canvasGridSize = canvas.scene.grid.size;
+    if (!canvasGridSize) return null;
+
+    const templateShape = isAmped ? "circle" : "rect";
+    const templateWidth = isAmped ? 0 : 5;
+    const templateDistance = isAmped ? 10 : 7.0710678118654755;
```
olilan1: Where did this number come from?

If there's a reason for it then as it's really unobvious, best to add a comment explaining it.

src/spells/redistributepotential.ts
```diff
@@ -0,0 +1,289 @@
...
+    const templateDirection = isAmped ? 0 : 45;
+
+    let flagName = "";
+    let flagSlug = "";
+    let color = "";
+    if (type === "steal-heat") {
+        flagName = "Redistribute Potential - Steal Heat";
+        flagSlug = "redistribute-potential-steal-heat";
+        color = "#4134d4";
```
olilan1: These colours appear at two places in the code, so if you later decide to change it, you'll have to remember to change it in both places.

Suggest defining a constant at the top of the file for each of the colour codes instead.

src/chatautobuttons.ts
```diff
@@ -54,20 +62,38 @@ export function addAutoButtonIfNeeded(message: ChatMessagePF2e, html: JQuery<HTM
...
+            const parentDiv = templateButton.closest(`div${divLookup}`);
+            const button = $('<button type="button">' + buttonSpec.label + '</button>');
+            button.on("click", function() {
+                buttonSpec.function(token, message);
+            });
+            parentDiv.after(button);
+            parentDiv.remove();
+        }
+    }
+}
```
olilan1: This is nice, appreciate you making it generic and reusable like this!

There's a little bit of duplication between the two functions in this file now, but no big deal.

src/spells/redistributepotential.ts
```diff
@@ -0,0 +1,293 @@
...
+    const concentrateHeatLocation = await selectLocation("concentrate-heat", stealHeatLocation, isAmped);
+    // If cancelled, clean up previous template
+    if (!concentrateHeatLocation) {
+        stealHeatTemplate.delete();
+        return;
+    }
+
+    // Create Concentrate Heat template and get target tokens
+    const concentrateHeatTemplateData = createCustomTemplateData(concentrateHeatLocation, "concentrate-heat", isAmped);
+    if (!concentrateHeatTemplateData) return;
```
olilan1: think this null check isn't needed any more

olilan1 (approved):

PR #80, opened 2025-11-09
Conservation of Energy

mustangsami:


src/conservationofenergy.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    // Check if a relevant spell
+    const prefix = "item:";
+    const isRelevantSpell = options.some(o => o.startsWith(prefix) 
+        && SPELL_SLUGS.includes(o.substring(prefix.length))
+    );
+
+    // Check if mind shift action with add remove energy enabled
+    const isMindShiftWithAddRemoveEnergy = options.includes("item:trait:mindshift") 
+        && options.includes("mindshift:add-remove-energy");
```
olilan1: This still doesn't work correctly, I'm afraid. If the Mindshift box is checked, but the Mindshift toggle is not set on the ability (e.g. if Psi Burst is used with the standard Bludgeoning damage), then the switch still happens. I guess this is because technically the ability always has the Mindshift trait, even if it's not being mindshifted.

src/conservationofenergy.ts
```diff
@@ -0,0 +1,107 @@
...
+    const payload = {
+        _id: oscillatingWaveFeat.id,
+        "system.rules": clonedRulesArray,
+    };
+
+    // Update the item on the actor
+    await psychicActor.updateEmbeddedDocuments("Item", [payload]);
+
+    // Animate energy switch for visual display of change
+    if (!message.token || !message.token.object) return;
```
olilan1: This could be a bit more concisely written as:

`if (!message.token?.object) return;`

olilan1 (commented):
This is a great feature, thanks for doing this!

Still a small issue to fix though, I think...

mustangsami:
I screwed something up but fixed it so ignore that there are two commits and just look at the last one.

src/conservationofenergy.ts
olilan1: I realised another issue with triggering this on damage rolls: it won't be right for spells with a duration. Namely _fire shield_, _flame vortex_, and _fiery body_, where the damage can occur after casting.

PR #87, opened 2025-11-15
Manifest Eidolon functionality and animation

mustangsami:


src/actions/manifesteidolon.ts
```diff
@@ -0,0 +1,216 @@
...
+    // Find eidolon actor associated with the summoner
+    const eidolonActor = getEidolonActor(summonerActor);
+    if (!eidolonActor) return;
+
+    // Check if that eidolon already has a token on the current scene
+    const eidolonTokens = getEidolonTokenOnCurrentScene(eidolonActor);
+    if (!eidolonTokens) {
+        const selectedEidolonManifestLocation = await selectEidolonManifestLocation(summonerToken, eidolonActor);
+        if (!selectedEidolonManifestLocation) return;
+        createChatMessageForGMClientDeManifest(summonerToken, eidolonActor, selectedEidolonManifestLocation);
```
olilan1: Should this function be called `createChatMessageForGMClientManifest`? Currently you have two functions with almost exactly the same name, just with the M capitalised on one of them

src/actions/manifesteidolon.ts
```diff
@@ -0,0 +1,216 @@
...
+        return;
```
olilan1: This return is not needed

src/actions/manifesteidolon.ts
```diff
@@ -0,0 +1,216 @@
...
+        content: messageContent,
+        whisper: ChatMessage.getWhisperRecipients("GM").map(u => u.id),
+        speaker: ChatMessage.getSpeaker({ actor: summonerToken.actor }),
+        flags: {
+            [MODULE_ID]: {
+                eidolonTokenId: eidolonToken.id,
+                type: "custom-demanifest-eidolon"
+            }
+        }
+    });
```
olilan1: Are these chat messages to the GM desirable, or are you just creating them as a way to run something on the GM's machine? If the latter, using socketlib might be cleaner, rather than showing an unnecessary chat message for this.

We could try to create a wrapper around socketlib if it's not convenient enough to use, I don't remember exactly how it works.

PR #91, opened 2025-11-18
BG3 Style Force Barrage

mustangsami:


src/spells/forcebarrage.ts
```diff
@@ -0,0 +1,337 @@
...
+        const templateData = await startCrosshairsTargetSelection(token, remainingShards);
+        if (!templateData && targets.length === 0) {
+            ui.notifications.error("Force barrage selection cancelled.");
+            return;
+        } else if (!templateData && targets.length > 0) {
+            targets.pop();
+            remainingShards++;
+            ui.notifications.warn("Last selected target removed.");
+            continue;
+        }
```
olilan1: You're checking `!templateData` twice. I think this would be a little clearer as:

``` TS
        if (!templateData) {
            if (targets.length < 1) {
                ui.notifications.error("Force barrage selection cancelled.");
                return;
            }
            targets.pop();
            remainingShards++;
            ui.notifications.warn("Last selected target removed.");
            continue;
        }
```

src/spells/forcebarrage.ts
```diff
@@ -0,0 +1,337 @@
...
+    // determine how many missiles the caster will have based on casting rank
+    const castRank = message.flags.pf2e.origin?.castRank;
+    if (!castRank) return;
+    const numberOfShards = numberOfActions * Math.ceil(castRank / 2)
+
+    // Ask player to designate shards to the tokens on the board
+    const targets = [];
+    let remainingShards = numberOfShards;
+    do  {
+        const templateData = await startCrosshairsTargetSelection(token, remainingShards);
```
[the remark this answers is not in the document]
mustangsami: Are you sure? The object type that is returned from crosshairs is type TemplateData. Happy to rename it, but it kinda _is_ a templateData....

src/spelldamageroll.ts
```diff
@@ -0,0 +1,125 @@
+import { SpellPF2e, TokenPF2e } from "foundry-pf2e";
+
+export type FormulaOverride = {_formula: string, formula: string, originalFormula: string}
```
olilan1: Per my other suggestion, I think this shouldn't be needed any more.

If we do need something like this, I'd say don't name things starting with an underscore, as that has a special meaning (it's for internal things). Better to give these more descriptive names - just looking at this type, it's not clear what the difference between `_formula`, `formula`, and `originalFormula` is.

src/spelldamageroll.ts
```diff
@@ -0,0 +1,125 @@
...
+    const validModifers = [];
+    if (!modifiers || modifiers.length === 0) return;
+    for (const modifier of modifiers) {
+        if (modifier.enabled) {
+            validModifers.push(modifier);
+        }
+    }
+    return validModifers;
+
+}
```
olilan1: Probably not needed any more if you apply my other suggestions

src/spelldamageroll.ts
```diff
@@ -0,0 +1,125 @@
+import { SpellPF2e, TokenPF2e } from "foundry-pf2e";
+
+export type FormulaOverride = {_formula: string, formula: string, originalFormula: string}
+type PF2ETraitKey = keyof typeof CONFIG.PF2E.traitsDescriptions;
+
+export async function rollSpellDamage(spellItem: SpellPF2e, targetToken: TokenPF2e,
```
olilan1: If this is potentially useful for other things later, it might be better to make it take an array of  tokens for the targets, rather than just one token.

src/utils.ts
```diff
@@ -316,4 +318,31 @@ export function getTokensOnCurrentSceneForActor(actor: ActorPF2e): TokenDocument
     const tokens = currentScene.tokens.filter(t => t.actorId === actor.id);
 
     return tokens;
+}
+
+export async function getTokensAtLocation(location: Point, includeHidden?: boolean): Promise<TokenPF2e[]> {
```
olilan1: It's not obvious at first glance that this excludes certian token types, so might be good to document that with a JSDoc comment. Something like this:

``` TS
/**
 * Returns all tokens at a given location, except for loot and party tokens.
 */
```

src/spelldamageroll.ts
```diff
@@ -0,0 +1,125 @@
+import { SpellPF2e, TokenPF2e } from "foundry-pf2e";
+
+export type FormulaOverride = {_formula: string, formula: string, originalFormula: string}
+type PF2ETraitKey = keyof typeof CONFIG.PF2E.traitsDescriptions;
+
+export async function rollSpellDamage(spellItem: SpellPF2e, targetToken: TokenPF2e, 
+    casterToken: TokenPF2e, formulaOverride?: FormulaOverride) {
+
+    const spellDamage = await spellItem.getDamage();
+    if (!spellDamage) return;
```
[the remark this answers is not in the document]
mustangsami: There was something weird going on with the object when I tried to pass it as SpellDamage. Not sure if you've given it a go, but it was complaining and I couldn't work out why that was. The only way I could get it to behave was to just call it again when I needed it.

PR #92, opened 2025-11-18
HuntPrey clean up to use Sockets, fixed Enjoy the Show ts errors

mustangsami:


src/actions/enjoytheshow.ts
```diff
@@ -1,36 +1,38 @@
...
 import { checkIfProvidesPanache } from "../effects/panache.ts";
 import { delay } from "../utils.ts";
 
+type OUTCOMES = "criticalSuccess" | "success" | "criticalFailure" | "failure";
+
 export function editEnjoyTheShowSkillRollIfNeeded(
     chatMessagePF2e: ChatMessagePF2e, html: JQuery<HTMLElement>) {
     const rollOptions = chatMessagePF2e.flags.pf2e.origin?.rollOptions;
     if (!rollOptions) return;
     if (!rollOptions.includes("origin:item:slug:enjoy-the-show")) return;
```
olilan1: While you're here, could remove line 10 and add a question mark after `rollOptions` in line 11

src/actions/huntprey.ts
```diff
@@ -1,105 +1,109 @@
...
-    const allowedNumberOfTargets = await calculateNumberOfTargetsRangerCanHunt(ranger);
-    if (rangerTargets.size > allowedNumberOfTargets) {
-        return;
+import { ActorPF2e, ChatMessagePF2e, ScenePF2e, TokenDocumentPF2e } from "foundry-pf2e";
+import { ImageFilePath } from "foundry-pf2e/foundry/common/constants.mjs";
+import { AUTO_HUNT_PREY, getSocket } from "../sockets.ts";
+
+export async function runHuntPreyAsGM(rangerUuid: string, targetUuids: string[]) {
+
+    const ranger = fromUuidSync(rangerUuid) as ActorPF2e;
```
olilan1: For consistency, `fromUuidSync<ActorPF2e>(rangerUuid)`

src/actions/huntprey.ts
```diff
@@ -1,105 +1,109 @@
...
-
-        if (token.actor === actor) continue;
-
-        const targetActor = token.actor;
-        if (!targetActor) continue; 
-
-        const items = await targetActor.items.contents;
-        const huntPreyEffect = items.find(item => 
+async function removeHuntPreyFromOtherTokens(hunterActor: ActorPF2e) {
+    const oldHuntedPreyTokenUuids = await hunterActor.getFlag("samioli-module", "huntedPreyTokenUuids")
```
olilan1: semicolon

src/actions/huntprey.ts
```diff
@@ -1,105 +1,109 @@
...
+    if (!oldHuntedPreyTokenUuids) return
```
olilan1: I think this line is unnecessary, because if it's null then isArray(blah) will be false

src/sockets.ts
```diff
@@ -1,10 +1,12 @@
+import { runHuntPreyAsGM } from "./actions/huntprey.ts";
 import { demanifestEidolonAsGM, manifestEidolonAsGM } from "./actions/manifesteidolon.ts";
 import { MODULE_ID } from "./utils.ts";
 
 let socket: SocketlibSocket | undefined;
 
 export const MANIFEST_EIDOLON = "manifestEidolon";
 export const DEMANIFEST_EIDOLON = "demanifestEidolon";
+export const AUTO_HUNT_PREY = "autoHuntPrey";
```
olilan1: Might be better for this name to be closer to the name of the function it calls.

So `RUN_HUNT_PREY`, or if you went with my earlier suggestion, `REMOVE_AND_APPLY_HUNT_PREY`

src/actions/huntprey.ts
mustangsami: I've done this but it actually then seems to complain about `ranger` on lines 16 and 17. Is it doing something marginally different?

src/actions/huntprey.ts
```diff
@@ -1,105 +1,109 @@
...
     }
 
-    const allowedNumberOfTargets = await calculateNumberOfTargetsRangerCanHunt(ranger);
+    const allowedNumberOfTargets = getNumberOfTargetsRangerCanHunt(ranger!);
     if (targets.size > allowedNumberOfTargets) {
         ui.notifications.error(`Please select a maximum of ${allowedNumberOfTargets} creature(s) to hunt`);
         return;
     }
+
+    getSocket().executeAsGM(AUTO_HUNT_PREY, ranger?.uuid, Array.from(targets).map(t => t.document.uuid));
```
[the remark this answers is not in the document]
mustangsami: My typescript doesn't seem to agree with that... and just moved the error up.

src/actions/huntprey.ts
olilan1: It's because `fromUuidSync` might not find anything, so it could be null.

By using `as ActorPF2e`, you're saying it's not null, so you're hiding that possible error.

src/sockets.ts
```diff
@@ -1,12 +1,14 @@
...
+import { startDazzlingDisplayAsGM } from "./actions/dazzlingdisplay.ts";
 import { demanifestEidolonAsGM, manifestEidolonAsGM } from "./actions/manifesteidolon.ts";
 import { MODULE_ID } from "./utils.ts";
 
 let socket: SocketlibSocket | undefined;
 
 export const MANIFEST_EIDOLON = "manifestEidolon";
 export const DEMANIFEST_EIDOLON = "demanifestEidolon";
-export const AUTO_HUNT_PREY = "autoHuntPrey";
+export const  REMOVE_AND_APPLY_HUNT_PREY = "removeAndApplyHuntPrey";
```
olilan1: nit: extra space here

src/sockets.ts
```diff
@@ -18,5 +20,6 @@ export function registerSocket() {
     // Register the name of the function that you want to run as the GM
     socket.register(MANIFEST_EIDOLON, manifestEidolonAsGM);
     socket.register(DEMANIFEST_EIDOLON, demanifestEidolonAsGM);
+    socket.register( REMOVE_AND_APPLY_HUNT_PREY, removeAndApplyHuntPreyAsGM);
```
olilan1: extra space here too

PR #94, opened 2025-11-23
Ghostly-carrier

mustangsami:


src/spells/ghostlycarrier.ts
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    actorToCreate.folder = folder!.id;
+
+    const createdActor = await Actor.create(actorToCreate)!;
+    if (!createdActor) return;
+
+    const updateData = {
+        // Update Attributes and Saves to match caster
+        "system.attributes.ac.value": casterActor.attributes?.ac?.value ?? 10,
```
olilan1: Question makr not needed after `attributes` I think

src/spells/ghostlycarrier.ts
```diff
@@ -0,0 +1,213 @@
...
+    // Get caster location and add it Ghostly Carrier TokenDocument
+    const x = casterTokenDocument.x;
+    const y = casterTokenDocument.y;
+    const tokenDocument = await createdActor.getTokenDocument({ x, y }) as TokenDocumentPF2e;
+
+    // Create TokenData so the token can be created on the canvas
+    const ghostlyCarrierTokenData = tokenDocument.toObject();
+
+    // So that the Ghostly Carrier token doesn't appear below the caster
+    const sort = casterTokenDocument.sort
```
olilan1: semicolon

Also you only use this once, so you could just do:

`ghostlyCarrierTokenData.sort = casterTokenDocument.sort + 1;`

src/spells/ghostlycarrier.ts
```diff
@@ -0,0 +1,213 @@
...
+
+}
+
+export async function deleteGhostlyCarrierTokenFromEffect(item: ItemPF2e) {
+    
+    if (item.system.slug !== "samioli-ghostly-carrier") return;
+    const tokenDocument = await getGhostlyCarrierTokenDocumentFromItem(item);
+    if (!tokenDocument) return;
+
+    getSocket().executeAsGM(GHOSTLY_CARRIER_DELETE, tokenDocument.uuid);
```
olilan1: When you do this, you're going from the item, to the token uuid, to the token document, then to the uuid again. So you're getting the token document unnecessarily.

It may be better to split `getGhostlyCarrierTokenDocumentFromItem` into two functions, one which gets the uuid, and one which gets the TokenDocument.

src/spells/ghostlycarrier.ts
```diff
@@ -0,0 +1,213 @@
...
+                ghostlyCarrierTokenDocumentUUID: ghostlyCarrierTokenDocument.uuid
+            }
+        }
+    } as DeepPartial<EffectSource> as EffectSource;
+
+    return await addOrUpdateEffectOnActor(casterActor, ghostlyCarrierEffectData);
+
+}
+
+export async function deleteGhostlyCarrierTokenFromEffect(item: ItemPF2e) {
```
olilan1: "from effect" in the name here makes it sound like you're deleting the token that belongs to the effect, which I guess isn't what you meant.

I guess you meant "delete the ghostly carrier token given this effect", but the name was initially a bit confusing to me.

Perhaps `deleteGhostlyCarrierTokenOnEffectDeletion`

.gitignore
```diff
@@ -1,3 +1,6 @@
 packs/macros
 node_modules
-dist
\ No newline at end of file
+dist
+**/LOCK
+**/LOG
```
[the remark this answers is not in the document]
olilan1: Also I'm seeing files called MANIFEST-[some number] and also a file called CURRENT appear in the packs directory - should they be added to the repo, or should they be ignored?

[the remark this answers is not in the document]
mustangsami: Yes to LOG.old

As far as I ynderstand MANIFEST and CURRENT are meant to be in the repo.

If we can get the CLI pack/unpack thing working then they won't be needed but I might need your help for that.

PR #97, opened 2025-12-06
shifting poc

mustangsami:


src/actions/shifting.ts
```diff
@@ -0,0 +1,302 @@
...
+import { ApplicationRenderOptions } from "foundry-pf2e/foundry/client/applications/_module.mjs";
+import { sendBasicChatMessage } from "../utils.ts";
+const { ApplicationV2, HandlebarsApplicationMixin } = foundry.applications.api;
+
+export async function displayShiftingWeaponDialog(token: TokenPF2e, message: ChatMessagePF2e) {
+
+    // Check for slug: "origin:item:activation-shift-weapon"
+    const rollOptions = message.flags.pf2e.origin?.rollOptions;
+
+    if (!rollOptions?.includes("origin:item:activation-shift-weapon")) return;
```
olilan1: Don't need to check this here any more

mustangsami:
I think I fixed the issue where a 2H weapon would be treated as 1H after it's been shifted once.

src/actions/shifting.ts
```diff
@@ -0,0 +1,311 @@
...
+    // Check how many hands the item is (this determines what it can shift into)
+
+    const baseWeaponUsage = currentWeapon.getFlag("samioli-module", "originalBaseWeaponUsage");
+
+    let isTwoHanded = false;
+    if (baseWeaponUsage) {
+        isTwoHanded = baseWeaponUsage === "held-in-two-hands";
+    } else {
+        isTwoHanded = currentWeapon.system.usage.value === "held-in-two-hands";
+    }
```
olilan1: Can we use `system.usage.hands` instead of `system.usage.value`, as `hands` seems to be just a number `1` or `2`

src/actions/shifting.ts
```diff
@@ -0,0 +1,311 @@
...
+ */
+function extractWeaponFromContent(content: string): WeaponPF2e | undefined {
+    const uuidMatch = content.match(/data-uuid="([^"]+)"/);
+    if (!uuidMatch) return undefined;
+
+    const uuid = uuidMatch[1];
+
+    const item = fromUuidSync(uuid);
+
+    if ((item as any)?.isOfType?.("weapon")) {
```
olilan1: I think you can get rid of the `any` and some of the weirdness here by making these lines do this:

``` TS
    const item = fromUuidSync(uuid) as ItemPF2e | null;

    if (item?.isOfType("weapon")) {
        return item as WeaponPF2e;
    }
```

mustangsami:
@olilan1 I think this is ready for a review again.

src/actions/shifting.ts
```diff
@@ -24,14 +19,13 @@ export async function displayShiftingWeaponDialog(token: TokenPF2e, message: Cha
...
 
     let isTwoHanded = false;
-    if (baseWeaponUsage) {
-        isTwoHanded = baseWeaponUsage === "held-in-two-hands";
+    if (baseWeaponHands) {
+        isTwoHanded = baseWeaponHands === 2;
     } else {
-        isTwoHanded = currentWeapon.system.usage.value === "held-in-two-hands";
+        isTwoHanded = currentWeapon.system.usage.hands === 2;
     }
```
olilan1: These six lines (24-29) could be just:

```
let isTwoHanded = (baseWeaponHands ?? currentWeapon.system.usage.hands) === 2;
```

olilan1 (commented):
Added one more comment, but there were also a few unresolved ones from before

mustangsami:
Ok - I think that's all the comments now?

olilan1 (approved):

PR #99, opened 2026-01-12
Antagonize could be stacked multiple times on a token

mustangsami:


src/actions/antagonize.ts
```diff
@@ -60,6 +68,15 @@ async function applyAntagonizedEffect(targetToken: TokenPF2e, antagonizerToken:
     
 }
 
+function checkIfAntagonizedByDemoralizer(actorToken: TokenPF2e, demoralizerToken: TokenPF2e) {
```
olilan1: As this returns a boolean and is used in an `if` statement, it'd read better if it was called `isAntagonizedByDemoralizer`

src/actions/antagonize.ts
```diff
@@ -60,6 +68,15 @@ async function applyAntagonizedEffect(targetToken: TokenPF2e, antagonizerToken:
...
+    const antagonizedEffects = getActorAntagonizedEffects(actorToken.actor!);
+    if (antagonizedEffects.length === 0) return false;
```
olilan1: I think this line isn't needed, because `some` will just return false if the thing is empty

src/actions/antagonize.ts
```diff
@@ -60,6 +68,15 @@ async function applyAntagonizedEffect(targetToken: TokenPF2e, antagonizerToken:
...
+    if (antagonizedEffects.some(effect => effect.flags.samioli?.antagonizerTokenId === demoralizerToken.id)) {
+        return true;
+    }
+    return false;
```
olilan1: no need to have the `if` with `return true` and then `return false`. You can just do:

```
return antagonizedEffects.some(effect => effect.flags.samioli?.antagonizerTokenId === demoralizerToken.id);
```

olilan1 (approved):

PR #101, opened 2026-01-16
Ghostly Carrier without the Compendium

mustangsami:
…it's empty


src/spells/ghostlycarrier.ts
```diff
@@ -53,7 +100,9 @@ export async function summonGhostlyCarrierAsGM(casterTokenUuid: string) {
         "system.saves.reflex.value": casterActor.saves?.reflex?.mod ?? 0,
         "system.saves.will.value": casterActor.saves?.will?.mod ?? 0,
         // set ownership to match caster
-        ownership: casterActor.ownership
+        ownership: casterActor.ownership,
+        // put the actor into the folder we created earlier
+        folder: folder!.id
     };
 
     await createdActor.update(updateData);
```
olilan1: This is a second database operation that also has to be `await`ed - it would be better to include these values in the object when you create it.

One way you could do this is replace your GHOSTLY_CARRIER_DATA constant with a `buildGhostlyCarrierData` function that takes the `castorActor` and `folder`. That function can then create and return the data, and you can set the saves and other things in the data so it's all there when you create the actor.

olilan1 (approved):

PR #102, opened 2026-02-04
readded packs/actors to the gitignore since the ghostlycarrier change…

mustangsami:
… - missed this in the last commit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
 packs/macros
-packs/actors
 node_modules
 dist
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -1,8 +1,4 @@
 packs/macros
+packs/actors
```
olilan1: I don't think `packs/actors` was in the file before, is it needed now? (Do you just need to delete the contents of the directory?)

mustangsami:
Ready for re-review

olilan1 (approved):

PR #103, opened 2026-02-10
Snares

mustangsami:


src/actions/snare.ts
```diff
@@ -0,0 +1,299 @@
...
+function generateSnareScript(deployerUuid: string, snareId: string, itemUuid: string) {
+
+    const script = `
+if (!game.user.isGM) return;
+const snareId = "${snareId}";
+const itemUuid = "${itemUuid}";
+const deployerUuid = "${deployerUuid}";
+const token = event.data.token;
+const myApi = game.modules.get("samioli-module").api;
+myApi.handleSnareRegionEnter(snareId, itemUuid, deployerUuid, token);
```
olilan1: nit: it'd be a bit more concise to remove lines 102-104 and just do:

``` TS
myApi.handleSnareRegionEnter(${snareId}, ${itemUuid}, ${deployerUuid}, token);
```

src/actions/snare.ts
```diff
@@ -0,0 +1,299 @@
...
+`;
+    return script;
+}
+
+export async function triggerSnare(snareId: string, itemUuid: string, deployerUuid: string, token: TokenDocumentPF2e) {
+
+    animateSnareTrigger(token);
```
olilan1: This is an async function, so add `await`?

src/actions/snare.ts
```diff
@@ -0,0 +1,299 @@
...
+    const deployer = fromUuidSync(deployerUuid) as TokenDocumentPF2e;
+
+    createSnareTriggeredChatMessage(deployer, token, itemUuid, snareId);
```
olilan1: `await` here too?

src/actions/snare.ts
```diff
@@ -0,0 +1,299 @@
...
+    const snare = canvas.scene?.regions.find((region) => region.getFlag("samioli-module", "snareId") === snareId);
+    if (!snare) return;
+    await snare.delete();
+}
+
+async function createSnareTriggeredChatMessage(deployer: TokenDocumentPF2e, triggerer: TokenDocumentPF2e, 
+    itemUuid: string, snareId: string) {
+
+    const content = `
+        <p>${triggerer.name} stepped on my snare!</p>
```
olilan1: Can we check if the token is hidden by the GM, and if it is say "A creature" instead of its name?

olilan1: Not sure, but would it be easier for the GM if this said "stepped on [deployer's name]'s snare"?

src/actions/snare.ts
```diff
@@ -0,0 +1,299 @@
...
+    if (message.flags['samioli-module']?.snareId) {
+        swapButtons("remove-snare", AUTO_SWAP_BUTTONS_CONSUMABLES, '.card-buttons', token, message, html);
+        return;
+    }
+    const slug = rollOptions.find(item => item.startsWith(SNARE_PREFIX))?.slice(SNARE_PREFIX.length);
+    if (!slug) return;
+    swapButtons(slug, AUTO_SWAP_BUTTONS_CONSUMABLES, '.card-buttons', token, message, html);
+    return; 
+
+}
```
olilan1, line 154: Can we do this in the chatautobuttons file along with the other auto buttons rather than having this here? It'd be neater that way and would mean we could add other auto buttons for consumables later if needed

olilan1 (commented):
Haven't reviewed everything yet but here's what I've got for now

src/actions/snare.ts
mustangsami: So it's actually spoken by the player, so it has their token icon and name in the chat box itself, so it's pretty clear I thought... Happy to change but it's basically there.

src/actions/snare.ts
mustangsami: So originally I did have it in there, the issue is that it's _so_ specific to snares at the moment that it basically acted like 

"Here is an easy helper to do buttons, but if you're a snare, do this entirely different thing and then do another different thing if it was already deployed"

Let me have another look at it, but the way I had it before I think you would have definitely not prefered.

src/actions/snare.ts
mustangsami: stupid question, but if I don't care to have this be synchronous, can't I leave it without an await?

src/actions/snare.ts
mustangsami: same here.

Maybe the real question is, does this function need to be async?

src/actions/snare.ts
olilan1: Not a stupid question. I guess you only need to await if you need the thing to finish before you move on, or if you care if the thing you're calling has an error. As this is an animation, maybe it doesn't matter here

src/actions/snare.ts
olilan1: If there's no `await` inside the function then I think it doesn't need to be async

src/actions/snare.ts
```diff
@@ -0,0 +1,299 @@
...
+
+export async function createSnareAsGM(location: Point, deployerUuid: string, snareId: string, 
+    itemUuid: string) {
+    const size = canvas.grid.size;
+    const script = generateSnareScript(deployerUuid, snareId, itemUuid);
+    
+    const regionData = {
+        name: "Snare",
```
olilan1: Can we include the deployer's name and the type of snare? Otherwise from the region list you can't tell which is which if there's more than one

src/actions/snare.ts
```diff
@@ -0,0 +1,299 @@
...
+        color: "#320101",
```
olilan1: Not sure if it would be better, but we could consider using the player's colour for this?

src/actions/snare.ts
olilan1: Yeah, we definitely don't want snare-specific code in chatautobuttons. The trick would be to find a way to add what you want in a generic way so we can use it for future use cases.

So here there are two cases, one where the message has a snareId, and one where it doesn't.

Perhaps we could add a new field to the `ButtonSpec` which is a condition on the message.

Something like this:

``` TS
condition?: (message: ChatMessagePF2e) => boolean;
```

And then the button specs for the two snare buttons becomes:

``` TS
   "snare": {
        label: "Deploy Snare!",
        function: deploySnare,
        condition: (message: ChatMessagePF2e) => message.flags['samioli-module']?.snareId === undefined
        buttonToReplace: USE_BUTTON_CONSUMABLE
    },
    "snare": {
        label: "Remove Snare?",
        function: removeSnare,
        condition: (message: ChatMessagePF2e) => message.flags['samioli-module']?.snareId !== undefined
        buttonToReplace: USE_BUTTON_CONSUMABLE
    }
```

... and then add code to check the new condition in `addMatchingButtons` and `swapButtons`

Oh, I guess there's one more thing needed, because for this you're matching on *category*, not on *slug*. I guess you define these as a separate set of button specs, e.g. `AUTO_SWAP_BUTTONS_CATEGORIES`, and add some code in `addAutoButtonIfNeeded` to get the category rather than the slug, and use that (similar to what you did on lines 181-182)

src/actions/snare.ts
```diff
@@ -0,0 +1,299 @@
...
+        content: content,
+        button_label: "Trigger Snare!",
+        params: [itemUuid, snareId, triggerer.uuid]
+    });
+}
+
+export async function addSnareToChatAndTarget(itemUuid: string, snareId: string, triggererUuid: string) {
+    const item = fromUuidSync(itemUuid) as ConsumablePF2e;
+    const triggerer = fromUuidSync(triggererUuid) as TokenDocumentPF2e;
+    const tokenUuid = triggerer.uuid;
```
olilan1: Don't you already have this uuid as `triggererUuid`?

src/actions/snare.ts
```diff
@@ -0,0 +1,299 @@
...
+    animateSnareTrigger(token);
+    const deployer = fromUuidSync(deployerUuid) as TokenDocumentPF2e;
+
+    createSnareTriggeredChatMessage(deployer, token, itemUuid, snareId);
+}
+
+async function removeSnare(_token: TokenPF2e, message: ChatMessagePF2e) {
+    const snareId = message.flags['samioli-module']?.snareId;
+    if (!snareId) return;
+    getSocket().executeAsGM(REMOVE_SNARE, snareId);
```
olilan1, line 109: Either add `await` here or remove `async` from the function.

src/actions/snare.ts
```diff
@@ -0,0 +1,299 @@
...
+function getSnareAtLocation(location: Point) {
+    return canvas.scene?.regions.find((region) => {
+        const snareId = region.getFlag("samioli-module", "snareId");
+        return !!snareId && region.x === location.x && region.y === location.y;
+    });
+}
+
+function getTokenAtLocation(location: Point) {
+    return canvas.tokens?.placeables.find((token) => token.x === location.x && token.y === location.y);
+}
```
olilan1: Would it be better to use the existing `getTokensAtLocation` function in utils? That handles some cases like large tokens I think

src/actions/snare.ts
```diff
@@ -0,0 +1,299 @@
...
+            lineJoin: "bevel",
+            strokeThickness: 3,
+            wordWrap: true,
+            wordWrapWidth: 200
+        });
+
+    new Sequence()
+    .effect()
+        // @ts-expect-error offset is valid
+        .atLocation(token, {offset: {x:0, y:-100}})
```
olilan1: When I tried this, the "Click!" seems to appear near the token's starting location (i.e. where they were when they started moving). Should this location be relative to the snare location instead?

src/hooks.ts
```diff
@@ -24,9 +24,15 @@ import { oscillateEnergy } from "./conservationofenergy.ts";
...
 import { deleteGhostlyCarrierEffectFromCaster, deleteGhostlyCarrierTokenOnEffectDeletion, moveGhostlyCarrierToCaster } from "./spells/ghostlycarrier.ts";
 import { displayShiftingWeaponDialog } from "./actions/shifting.ts";
+import { samiOliModuleAPI } from "./api.ts";
+import { replaceButtonsForSnareMessages } from "./actions/snare.ts";
 
 Hooks.on("init", () => {
     registerSettings();
+    const module = game.modules.get("samioli-module");
+    if (module) {
+        (module as any).api = samiOliModuleAPI;
```
olilan1: This `any` will cause a red underline.

So what's happening here is you're adding a new "api" property on the Module, which did not exist before. So TS is rightly saying that "api" is not a thing it recognises.

I think you can fix this by telling it that `module` here is a Module with a new `api` thing attached. Like this:

``` TS
(module as Module & { api: typeof samiOliModuleAPI }).api = samiOliModuleAPI;
```

Or you could put the new type on the line before:

``` TS
const module: Module & { api: typeof samiOliModuleAPI } = game.modules.get("samioli-module")!;
module.api = samiOliModuleAPI;
```

src/actions/snare.ts
mustangsami: So I picked this colour so that it wasn't insanely bright on the board. If the player's colour is super bright it's kinda jarring. But we could do that instead.

olilan1 (approved):

PR #109, opened 2026-04-06
Update create db script

olilan1:
Updates the create db script to read from a directory containing one directory per sound set, each containing Attack, Hurt, and Death subdirectories of sound files.

### DIFF
--- a/scripts/create-sound-db.js
+++ b/scripts/create-sound-db.js
@@ -1,81 +1,78 @@
-const fs = require('fs');
-const path = require('path');
+/**
+ * This script generates a JSON database of sound files mapped by sound set and category.
+ * 
+ * USAGE:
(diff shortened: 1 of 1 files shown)


mustangsami:
Why is this in SamiOli Module?

olilan1:
> Why is this in SamiOli Module?

I guess just legacy, this is where the previous script was so I just edited it in place. I guess I could move this to pf2e-creature-sounds

mustangsami:
Ah ok, I don't mind either way. You tell me what you'd prefer to do.

olilan1:
> Ah ok, I don't mind either way. You tell me what you'd prefer to do.

Maybe submit this here for now, and we can move all the scripts later if we want.

mustangsami (approved):

PR #112, opened 2026-04-10
Added a Macro to trigger the shifting weapon dialog

mustangsami:
Added a Macro to trigger the shifting weapon dialog

Changed the naming of the Shifted weapon so AA does the correct animation

Highlight the Original Weapon Form in the Shifting Weapon UI

### DIFF
--- a/src/actions/shifting.ts
+++ b/src/actions/shifting.ts
@@ -2,27 +2,100 @@ import { ChatMessagePF2e, ItemPF2e, TokenPF2e, WeaponPF2e } from "foundry-pf2e";
 import { sendBasicChatMessage } from "../utils.ts";
 import { ShiftingWeaponApp } from "../ui/shiftingui.ts";
 
-export async function displayShiftingWeaponDialog(token: TokenPF2e, message: ChatMessagePF2e) {
+const { DialogV2 } = foundry.applications.api;
+
+export async function displayShiftingWeaponDialogFromActivationsModule(token: TokenPF2e, message: ChatMessagePF2e) {
 
     // Get the item from the message content
-    const currentWeapon = extractWeaponFromContent(message.content);
-    if (!currentWeapon) {
+    const weapon = extractWeaponFromContent(message.content);
+    if (!weapon) {
         ui.notifications.warn("This action acts on a weapon, but no weapon was found.");
         return;
     }
 
-    // Check that the item has the shifting rune
-    if (!hasShiftingRune(currentWeapon)) {
-        ui.notifications.warn(`The weapon ${currentWeapon.name} does not have a Shifting rune.`);
+    displayShiftingWeaponDialogForWeapon(token, weapon);
+}
+
+export async function displayShiftingWeaponDialogViaMacro(token: TokenPF2e) {
+    const weapon = await getHeldShiftingWeaponFromToken(token);
+    if (!weapon) return;
+    displayShiftingWeaponDialogForWeapon(token, weapon);
+}
+
+export async function getHeldShiftingWeaponFromToken(token: TokenPF2e): Promise<WeaponPF2e | null> {
+    if (!token.actor) return null;
+
+    // Filter for all held weapons
+    const heldWeapons = token.actor.itemTypes.weapon.filter(w => w.isEquipped);
+
+    // Determine if it has a shifting rune
+    const shiftingWeapons = heldWeapons.filter(w => hasShiftingRune(w));
+
+    if (shiftingWeapons.length === 0) {
+        ui.notifications.warn("No equipped weapon with a shifting rune was found.");
+        return null;
+    }
+
+    if (shiftingWeapons.length === 1) {
+        return shiftingWeapons[0];
+    }
+
+    // Handle multiple shifting weapons scenario
+    try {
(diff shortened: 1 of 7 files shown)


src/actions/shifting.ts
```diff
@@ -2,65 +2,144 @@ import { ChatMessagePF2e, ItemPF2e, TokenPF2e, WeaponPF2e } from "foundry-pf2e";
...
+    const weapon = extractWeaponFromContent(message.content);
+    if (!weapon) {
         ui.notifications.warn("This action acts on a weapon, but no weapon was found.");
         return;
     }
 
-    // Check that the item has the shifting rune
-    if (!hasShiftingRune(currentWeapon)) {
-        ui.notifications.warn(`The weapon ${currentWeapon.name} does not have a Shifting rune.`);
+    displayShiftingWeaponDialogForWeapon(token, weapon);
```
olilan1: As this function is `async`, should probably `await` here.

(It might not make a lot of difference in this case, but as you've declared `displayShiftingWeaponDialogFromActivationsModule` as async, it probably makes sense to await for this)

src/actions/shifting.ts
```diff
@@ -2,65 +2,144 @@ import { ChatMessagePF2e, ItemPF2e, TokenPF2e, WeaponPF2e } from "foundry-pf2e";
...
+}
+
+export async function displayShiftingWeaponDialogViaMacro(token: TokenPF2e) {
+    const weapon = await getHeldShiftingWeaponFromToken(token);
+    if (!weapon) return;
+    displayShiftingWeaponDialogForWeapon(token, weapon);
```
olilan1: `await` here too

src/actions/shifting.ts
```diff
@@ -2,65 +2,144 @@ import { ChatMessagePF2e, ItemPF2e, TokenPF2e, WeaponPF2e } from "foundry-pf2e";
...
+        ui.notifications.warn(`The weapon ${weapon.name} does not have a Shifting rune.`);
         return;
     }
 
+    // Set the original weapon flags if this is the first time shifting this weapon
+    if (!getOriginalBaseWeapon(weapon)) {
+        await setOriginalWeaponFlagsOnWeapon(weapon);
+    }
+
+    const originalBaseWeapon = getOriginalBaseWeapon(weapon);
```
olilan1: You are calling `getOriginalBaseWeapon` twice here in the case where the weapon doesn't have the base flags set yet.

Better to do this:

``` TS
    let originalBaseWeapon = getOriginalBaseWeapon(weapon);

    // Set the original weapon flags if this is the first time shifting this weapon
    if (!originalBaseWeapon) {
        await setOriginalWeaponFlagsOnWeapon(weapon);
        originalBaseWeapon = weapon.system.baseItem;
    }
```

src/actions/shifting.ts
```diff
@@ -2,65 +2,144 @@ import { ChatMessagePF2e, ItemPF2e, TokenPF2e, WeaponPF2e } from "foundry-pf2e";
...
 
-    const content = `${token.name} shifts their ${removeShiftingSuffix(currentWeapon.name)} into a ${selectedWeapon.name}.`;
-
     // Update the existing weapon to match stats of the selected weapon form
-    await updateWeaponStats(currentWeapon, selectedWeapon);
+    const isOriginalForm = await updateWeaponStats(weapon, selectedWeapon);
+
+    let content = "";
+    if (isOriginalForm){
+        content = `${token.name} shifts their ${extractOriginalNameFromShiftedSuffix(weapon.name)} into it's original form.`;
```
olilan1: `its`, not `it's`

olilan1 (commented):
Good stuff!

src/actions/shifting.ts
```diff
@@ -2,65 +2,145 @@ import { ChatMessagePF2e, ItemPF2e, TokenPF2e, WeaponPF2e } from "foundry-pf2e";
...
-
     // Update the existing weapon to match stats of the selected weapon form
-    await updateWeaponStats(currentWeapon, selectedWeapon);
+    const isOriginalForm = await updateWeaponStats(weapon, selectedWeapon);
+
+    let content = "";
+    if (isOriginalForm){
+        content = `${token.name} shifts their ${extractOriginalNameFromShiftedSuffix(weapon.name)} into its original form.`;
+    } else {
+        content = `${token.name} shifts their ${extractOriginalNameFromShiftedSuffix(weapon.name)} into a ${selectedWeapon.name}.`;
```
olilan1, line 114: Might it be better to save the original name in the same way that we save the original base weapon, to avoid having to extract it with a regex each time? Also would protect against anyone changing the name of the shifted weapon.

mustangsami: I did consider that but then the issue is if you ever _did_ want to change the name later down the line. There would be no way to overwrite the flag without doing some hackery in the console.

olilan1: Hmm, I guess the player might want to rename it, e.g. if the runes change.

It still seems a bit fragile, like it'll break if the player renames it but changes the format. Maybe that's ok for now

olilan1 (approved):

PR #114, opened 2026-04-10
Shifting weapon upgrades NEW PR

mustangsami:
This PR should be on top of Shifting-Rune-Macro

### DIFF
--- a/src/actions/shifting.ts
+++ b/src/actions/shifting.ts
@@ -1,4 +1,4 @@
-import { ActorPF2e, ChatMessagePF2e, ItemPF2e, TokenPF2e, WeaponPF2e } from "foundry-pf2e";
+import { ActorPF2e, CharacterPF2e, ChatMessagePF2e, ItemPF2e, MartialProficiency, TokenPF2e, WeaponPF2e } from "foundry-pf2e";
 import { sendBasicChatMessage } from "../utils.ts";
 import { ShiftingWeaponApp } from "../ui/shiftingui.ts";
 
@@ -203,7 +203,7 @@ function extractWeaponFromContent(content: string): WeaponPF2e | null {
 }
 
 function getActorAttackTrainedProficiencies(actor: ActorPF2e): ActorAttackTrainedProficiencies {
-    const attacks = (actor as any)?.system?.proficiencies?.attacks || {};
+    const attacks = (actor as CharacterPF2e)?.system?.proficiencies?.attacks || {};
     
     // Check if rank is 1 or more for simple and martial weapons
     const simple = (attacks.simple?.rank ?? 0) >= 1;
@@ -214,7 +214,7 @@ function getActorAttackTrainedProficiencies(actor: ActorPF2e): ActorAttackTraine
     // Check for advanced weapon proficiencies
     for (const [key, value] of Object.entries(attacks)) {
         if (key.startsWith("weapon-base-")) {
-            const rank = (value as any)?.rank ?? 0;
+            const rank = (value as MartialProficiency)?.rank ?? 0;
             if (rank >= 1) {
                 const slug = key.replace("weapon-base-", "");
                 advanced.push(slug);
--- a/src/actions/shifting.ts
+++ b/src/actions/shifting.ts
@@ -1,14 +1,8 @@
-import { ActorPF2e, CharacterPF2e, ChatMessagePF2e, ItemPF2e, MartialProficiency, TokenPF2e, WeaponPF2e } from "foundry-pf2e";
+import { CharacterPF2e, ChatMessagePF2e, ItemPF2e, TokenPF2e, WeaponPF2e } from "foundry-pf2e";
 import { sendBasicChatMessage } from "../utils.ts";
 import { ShiftingWeaponApp } from "../ui/shiftingui.ts";
 
 const { DialogV2 } = foundry.applications.api;
-
-export interface ActorAttackTrainedProficiencies {
-    simple: boolean;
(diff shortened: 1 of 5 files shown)


src/actions/shifting.ts
```diff
@@ -191,4 +200,31 @@ function extractWeaponFromContent(content: string): WeaponPF2e | null {
     }
 
     return null;
+}
+
+function getActorAttackTrainedProficiencies(actor: ActorPF2e): ActorAttackTrainedProficiencies {
+    const attacks = (actor as any)?.system?.proficiencies?.attacks || {};
```
olilan1: This `as any` will cause a red underline.

Think we might want `as CharacterPF2e` instead

src/actions/shifting.ts
```diff
@@ -191,4 +200,31 @@ function extractWeaponFromContent(content: string): WeaponPF2e | null {
...
+    // Check if rank is 1 or more for simple and martial weapons
+    const simple = (attacks.simple?.rank ?? 0) >= 1;
+    const martial = (attacks.martial?.rank ?? 0) >= 1;
+    
+    const advanced: string[] = [];
+    
+    // Check for advanced weapon proficiencies
+    for (const [key, value] of Object.entries(attacks)) {
+        if (key.startsWith("weapon-base-")) {
+            const rank = (value as any)?.rank ?? 0;
```
olilan1: I think this `as any` should be `as MartialProficiency`.

src/actions/shifting.ts
```diff
@@ -191,4 +200,31 @@ function extractWeaponFromContent(content: string): WeaponPF2e | null {
...
+    
+    // Check if rank is 1 or more for simple and martial weapons
+    const simple = (attacks.simple?.rank ?? 0) >= 1;
+    const martial = (attacks.martial?.rank ?? 0) >= 1;
+    
+    const advanced: string[] = [];
+    
+    // Check for advanced weapon proficiencies
+    for (const [key, value] of Object.entries(attacks)) {
+        if (key.startsWith("weapon-base-")) {
```
olilan1: Weapon proficiencies are more complicated than this, I'm afraid.

If you have Dwarven Weapon Familiarity for example, you get proficiency with all weapons with the dwarf trait, and treat martial ones as simple weapons and advanced ones as martial weapons. I think this adds a special entry in here that contains a predicate.

It's also possible to get proficiency in all advanced weapons from a given group with Advanced Weapon Training, or even with all Advanced weapons with Diverse Weapon Expert.

I'm not sure if there's an easy way to check what proficiency a character would have with a given weapon. The system has a [`getWeaponProficiencyRank`](https://github.com/foundryvtt/pf2e/blob/97a37f1fe60802a6cfb1626048341f925f7bd717/src/module/actor/character/helpers.ts#L140) function, but not sure we can call it.

src/ui/shiftingui.ts
```diff
@@ -82,6 +85,53 @@ export class ShiftingWeaponApp extends HandlebarsApplicationMixin(ApplicationV2)
                 }
             });
         });
+
+        // Search and filter functionality
```
olilan1, line 89: Would it be better to do this in `_prepareContext` instead? Otherwise we are still rendering the rows for all the weapons and then just making the filtered-out ones hidden. There's already a filter part in prepareContext, so could add more to it for this.

src/ui/shiftingui.ts
```diff
@@ -82,6 +85,53 @@ export class ShiftingWeaponApp extends HandlebarsApplicationMixin(ApplicationV2)
...
+
+            rows.forEach(row => {
+                const nameCell = row.querySelector(".weapon-name strong");
+                const traits = Array.from(row.querySelectorAll(".tags .tag")).map(t => t.textContent?.toLowerCase() || "");
+                const category = row.getAttribute("data-category") || "";
+                const slug = row.getAttribute("data-slug") || "";
+
+                const nameMatch = nameCell?.textContent?.toLowerCase().includes(query) ?? false;
+                const traitMatch = traits.some(t => t.includes(query));
+                const matchesSearch = nameMatch || traitMatch || query === "";
```
olilan1: Suggest breaking this out to a separate function that takes a weapon and returns whether it matches the search term or not.

src/actions/shifting.ts
mustangsami: Oh.... errr... what then? Should I try and catch those edge cases or maybe I just remove this feature overall?

src/ui/shiftingui.ts
mustangsami: If we do that then we have to re-render the UI whenever the filter or search is done, which is what I'm not doing because it would be sluggish when filtering. It makes sense for the 1H/2H filter as that weapon can _never_ be a 2H one if it's originally 1H.

olilan1: But this code you have here is in `_onRender`, so doesn't that mean you're already re-rendering the UI whenever the filter gets changed?

src/actions/shifting.ts
olilan1: Up to you if you want to try it. I think you'd have to basically copy that `getWeaponProficiencyRank` function and call it for each weapon.

If you do that I guess you may as well just take all of `actor.system.proficiencies.attacks` and use that rather than trying to create your own `ActorAttackTrainedProficiencies` as you are here.

src/ui/shiftingui.ts
olilan1: ok, I see how this is working now

olilan1 (approved):

PR #118, opened 2026-04-18
Add animation button for translocate and amped warp step

olilan1:
Also allows chat auto buttons to work if the card doesn't already contain buttons.

### DIFF
--- a/src/spells/translocate.ts
+++ b/src/spells/translocate.ts
@@ -1,4 +1,5 @@
 import { ChatMessagePF2e, CreaturePF2e, TokenPF2e } from "foundry-pf2e";
+import { getCollidableCallbacks } from "../utils.ts";
 
 const TELEPORT_SOUND = "sound/NWN2-Sounds/sfx_Teleportation.wav";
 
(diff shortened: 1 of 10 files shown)


src/spells/translocate.ts
```diff
@@ -0,0 +1,125 @@
...
+            .waitUntilFinished(-250)
+            .animation()
+            .teleportTo(destination) // Teleport to location
+            .snapToGrid()
+            .on(token)
+            .opacity(1)
+        .play()
+}
+
+async function selectDestination(token: TokenPF2e, maxRange: number, icon: string) {
```
mustangsami, line 121: Maybe worth telling the player what they are selecting? I usually put a an info or something for them to know what they're doing.

olilan1: done

olilan1:
Also added some other changes I think at the same time you were looking at this, to add sound and fix some things in the auto button code

src/spells/translocate.ts
```diff
@@ -0,0 +1,132 @@
...
+            .snapToGrid()
+            .opacity(1)
+        .play()
+}
+
+async function selectDestination(token: TokenPF2e, maxRange: number, icon: string) {
+    ui.notifications.info("Select destination for teleport.");
+    
+    return await Sequencer.Crosshair.show({
+        location: {
```
mustangsami: Add "wallBehavior: Sequencer.Crosshair.PLACEMENT_RESTRICTIONS.NO_COLLIDABLES" or else the player can select something off screen or behind a wall etc.

mustangsami (changes requested):
Looks awesome. Just maybe add that placement restriction and you're good to go.

src/spells/translocate.ts
olilan1: Done. I also added a helper function for the callbacks as we had those duplicated in lots of places.

mustangsami (approved):

PR #123, opened 2026-04-30
Implement the mood cloud Emote action

olilan1:
Token images for each of the three moods must be provided in Settings

### DIFF
--- a/src/actions/emote.ts
+++ b/src/actions/emote.ts
@@ -19,11 +19,16 @@ const MOOD_SETTINGS = {
     "Inscrutable": SETTINGS.MOOD_CLOUD_INSCRUTABLE_IMAGE
 } as const;
 
-export async function startEmote(token: TokenPF2e, _message: ChatMessagePF2e) {
+export async function startEmote(token: TokenPF2e, _message?: ChatMessagePF2e) {
     const actor = token.actor;
 
     if (!actor || actor.type !== "familiar") {
         ui.notifications.warn(`Emote must be performed by a familiar.`);
+        return;
+    }
+
+    if (!actor.isOwner) {
+        ui.notifications.warn(`You do not have permission to perform an emote with ${token.name}.`);
         return;
     }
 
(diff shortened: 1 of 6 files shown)


olilan1 (commented):
The automation is added as a chat button to the Emote action. This can be added to a familiar by adding the Mood Cloud ability from the PF2e Specific Familiars module.

src/actions/emote.ts
```diff
@@ -0,0 +1,180 @@
...
+    "Inscrutable": "deception"
+};
+
+const MOOD_SETTINGS = {
+    "Happy": SETTINGS.MOOD_CLOUD_HAPPY_IMAGE,
+    "Angry": SETTINGS.MOOD_CLOUD_ANGRY_IMAGE,
+    "Inscrutable": SETTINGS.MOOD_CLOUD_INSCRUTABLE_IMAGE
+} as const;
+
+export async function startEmote(token: TokenPF2e, _message: ChatMessagePF2e) {
```
mustangsami: Consider adding a way to trigger this via a macro that we can also add to the compendium. Similar to how we have two ways to initiate Shifting Weapon. Useful if the 3rd party module isn't installed etc.

src/actions/emote.ts
```diff
@@ -0,0 +1,180 @@
...
+                expiry: "turn-start"
+            },
+            rules: [
+                {
+                    key: "FlatModifier",
+                    selector: skillSlug,
+                    value: bonusValue,
+                    type: "circumstance",
+                    label: `Aided by ${familiarName}`,
+                    removeAfterRoll: true
```
mustangsami, line 150: It might be nice as a future improvement to have this be removed on the relevant roll. At the moment, any roll removes this.

mustangsami (approved):

src/actions/emote.ts
olilan1: OK, done

src/actions/emote.ts
olilan1: Oh, I didn't relaise that. Might be tricky as we'd have to have our mechanism to do this I guess. I added a TODO

PR #124, opened 2026-05-01
Shifting weapon improvements

olilan1:
- Eliminate loading time on first show of dialog
- Change weapon icon when shifting
- Move original form to the top of the list
- Search works on all text (e.g. you can search by "d4")
- Code style and efficiency improvements

### DIFF
--- a/src/actions/shifting.ts
+++ b/src/actions/shifting.ts
@@ -1,10 +1,23 @@
-import { CharacterPF2e, ChatMessagePF2e, ItemPF2e, TokenPF2e, WeaponPF2e } from "foundry-pf2e";
-import { sendBasicChatMessage } from "../utils.ts";
+import { ChatMessagePF2e, ItemPF2e, TokenPF2e, WeaponPF2e } from "foundry-pf2e";
+import { isCharacter, MODULE_ID, sendBasicChatMessage } from "../utils.ts";
 import { ShiftingWeaponApp } from "../ui/shiftingui.ts";
 
 const { DialogV2 } = foundry.applications.api;
 
-export async function displayShiftingWeaponDialogFromActivationsModule(token: TokenPF2e, message: ChatMessagePF2e) {
+const UUID_REGEX = /data-uuid="([^"]+)"/;
+const SHIFTED_NAME_REGEX = / \(shifted (.*)\)$/;
+
+const SHIFTING_FLAGS = {
+    BASE_WEAPON: "originalBaseWeapon",
+    BASE_HANDS: "originalBaseWeaponHands",
+    ORIGINAL_NAME: "originalWeaponName",
+    ORIGINAL_IMG: "originalWeaponImg"
+} as const;
+
+export async function displayShiftingWeaponDialogFromActivationsModule(
+    token: TokenPF2e, 
+    message: ChatMessagePF2e
+): Promise<void> {
 
     // Get the item from the message content
     const weapon = extractWeaponFromContent(message.content);
@@ -16,7 +29,7 @@ export async function displayShiftingWeaponDialogFromActivationsModule(token: To
     await displayShiftingWeaponDialogForWeapon(token, weapon);
 }
 
-export async function displayShiftingWeaponDialogViaMacro(token: TokenPF2e) {
+export async function displayShiftingWeaponDialogViaMacro(token: TokenPF2e): Promise<void> {
     const weapon = await getHeldShiftingWeaponFromToken(token);
     if (!weapon) return;
     await displayShiftingWeaponDialogForWeapon(token, weapon);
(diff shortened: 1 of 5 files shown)


src/actions/shifting.ts
```diff
@@ -40,166 +52,214 @@ export async function getHeldShiftingWeaponFromToken(token: TokenPF2e): Promise<
...
 
     // Check if the user selected something or closed the window
     if (!selectedWeapon) {
         ui.notifications.warn("Shifting cancelled.");
         return;
     }
 
+    // Prevent unnecessary database writes if selecting the exact same form
+    if (getWeaponBaseId(weapon) === getWeaponBaseId(selectedWeapon)) {
+        ui.notifications.info(`${token.name}'s weapon is already in that form.`);
```
mustangsami: Does this need to be displayed to the user?

src/ui/shiftingui.ts
```diff
@@ -35,78 +69,96 @@ export class ShiftingWeaponApp extends HandlebarsApplicationMixin(ApplicationV2)
...
-                if (hideUntrained) {
-                    matchesProficiency = row.getAttribute("data-is-proficient") === "true";
-                }
-
-                if (matchesSearch && matchesProficiency) {
-                    (row as HTMLElement).style.display = "";
-                } else {
-                    (row as HTMLElement).style.display = "none";
-                }
+                // Reading textContent is highly optimized compared to running DOM selectors. It
```
mustangsami: Can we clear out some of these clearly AI generated comments?

src/actions/shifting.ts
```diff
@@ -40,166 +52,214 @@ export async function getHeldShiftingWeaponFromToken(token: TokenPF2e): Promise<
...
+    return (currentWeapon.getFlag(MODULE_ID, SHIFTING_FLAGS.BASE_WEAPON) as string) ?? null;
 }
 
 function getOriginalWeaponName(currentWeapon: WeaponPF2e): string | null {
-    return currentWeapon.getFlag("samioli-module", "originalWeaponName") as string;
+    return (currentWeapon.getFlag(MODULE_ID, SHIFTING_FLAGS.ORIGINAL_NAME) as string) ?? null;
 }
 
-async function setOriginalWeaponNameFlag(weapon: WeaponPF2e) {
-    await weapon.setFlag("samioli-module", "originalWeaponName", weapon.name);
```
mustangsami, line 134: If I change the item name when it's in it's original form, it no longer applies the new name to it's shifted form.

Eg. Original name was: "Spellstriker Staff (Major)"

I changed it to "Spellstriker Staff" when in original form.

When I shifted it to another form, it became: "Chain Sword (shifted Spellstriker Staff (Major))"

Was this intentional?

src/actions/shifting.ts
```diff
@@ -40,166 +52,214 @@ export async function getHeldShiftingWeaponFromToken(token: TokenPF2e): Promise<
...
 
 /**
  * Updates the current weapon with specific traits from the selected weapon.
  */
-async function updateWeaponStats(preShiftedWeapon: WeaponPF2e, postShiftedWeapon: WeaponPF2e): Promise<boolean> {
-    const originalBaseWeapon = getOriginalBaseWeapon(preShiftedWeapon);
-    const isChangingToOriginalForm = originalBaseWeapon === postShiftedWeapon.system.baseItem;
-
-    // If we're shifting FROM the original form, save its current name first
-    if (originalBaseWeapon === preShiftedWeapon.system.baseItem) {
```
mustangsami, line 145: I think this needs to be reintroduced to solve the naming issue.

mustangsami (changes requested):

src/actions/shifting.ts
olilan1: This still happens, it was just moved down a bit (in this PR it's on line 215).

This was done so the name flag change could happen in the same update as other things, to avoid multiple DB updates.

But, the issue occurred because we are taking the original name from the flag all the time, and now we are setting the flag after that happens.

Fixed it by changing the code that gets the original name to just take the weapon's current name if it's in original form.

src/actions/shifting.ts
olilan1: No, a bug crept in when I moved the code that sets the original name flag lower down. Fixed (see other comment).

src/ui/shiftingui.ts
olilan1: OK yeah some of the comments were more justifying a change, rather than saying what the code actually did. Cleared out a lot of the comments.

src/actions/shifting.ts
olilan1: I guess not. But then, why even show the current form in the list?

I made a change to just filter out the current form so it doesn't appear.

olilan1 (commented):
PTAL

mustangsami (approved):

PR #130, opened 2026-05-07
Dancing blade spell implementation

olilan1:


src/spells/dancingblade-anim.ts
mustangsami, line 1: Might this be better suited within UI folder to go with the others?

src/spells/dancingblade-ui.ts
```diff
@@ -0,0 +1,456 @@
...
+}
+
+/**
+ * Plays a quick impact animation on the target for a strike or push action.
+ */
+export function playDancingBladeAttackAnimation(target: TokenPF2e, type: "strike" | "push") {
+    // prettier-ignore
+    new Sequence()
+        .effect()
+            .file(type === "strike" ? STRIKE_ANIM : PUSH_ANIM)
```
mustangsami: a small pause here would be good. It happens very quickly after the button has been pressed.

src/spells/dancingblade.ts
```diff
@@ -0,0 +1,808 @@
...
+    }
+    if (!selectedDamageType) return;
+
+    const statistic = spell.spellcasting?.statistic;
+    if (!statistic) {
+        ui.notifications.error("Could not find spellcasting statistic for Dancing Blade.");
+        return;
+    }
+
+    await playDancingBladeAttackAnimation(target, "strike");
```
mustangsami: await not needed

mustangsami (changes requested):

src/spells/dancingblade-anim.ts
olilan1: Not sure if that would be best, because most of this file is animations, and other animations are in this directory.

I renamed the file to `-anim` instead.

src/spells/dancingblade-ui.ts
olilan1: Done, also moved the animation to be after the roll so the pause doesn't delay the roll

src/spells/dancingblade.ts
olilan1: I think await is right, but the function it's calling isn't `async` and it should be. I changed it and other functions that play sequences to async.

mustangsami (approved):

PR #134, opened 2026-05-20
Update Start Of Turn Spells logic to workaround how PF2e-Dailies does…

mustangsami:
… spells via activations

### DIFF
--- a/src/startofturnspells.ts
+++ b/src/startofturnspells.ts
@@ -1,26 +1,26 @@
-import { ActorPF2e, CombatantPF2e, EffectSource, ItemPF2e, MeasuredTemplateDocumentPF2e, SpellPF2e, TokenPF2e } from "foundry-pf2e";
+import { ActorPF2e, CombatantPF2e, EffectSource, ItemPF2e, MeasuredTemplateDocumentPF2e, SpellPF2e, SpellSource, TokenPF2e } from "foundry-pf2e";
 import { addOrUpdateEffectOnActor, delay, sendBasicChatMessage } from "./utils.ts";
 import { getTemplateTokens, isTokenInTemplateArea, replaceTargets } from "./templatetarget.ts";
 
 const START_OF_TURN_SPELLS = [
-        'ash-cloud',
-        'field-of-life',
-        'sea-of-thought',
-        'visions-of-danger',
-        'flammable-fumes',
-        'earthquake',
-        'corrosive-muck',
-        'frozen-fog',
-        'control-sand',
-        'rust-cloud',
-        'petal-storm',
-        'antlion-trap',
-        'wall-of-fire',
-        'wall-of-virtue'    
+    'ash-cloud',
+    'field-of-life',
+    'sea-of-thought',
+    'visions-of-danger',
+    'flammable-fumes',
+    'earthquake',
+    'corrosive-muck',
+    'frozen-fog',
+    'control-sand',
+    'rust-cloud',
+    'petal-storm',
+    'antlion-trap',
+    'wall-of-fire',
+    'wall-of-virtue'
 ];
 
 export async function addEffectsToTokensInStartOfTurnTemplates(template: MeasuredTemplateDocumentPF2e) {
-    
+
     // @ts-expect-error slug is valid
     const spellSlug = template.flags.pf2e?.origin?.slug;
     const spellUuid = template.flags.pf2e?.origin?.uuid as string;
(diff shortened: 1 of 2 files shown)


src/startofturnspells.ts
```diff
@@ -23,12 +23,21 @@ export async function addEffectsToTokensInStartOfTurnTemplates(template: Measure
...
-
     const spellUuid = template.flags.pf2e?.origin?.uuid as string;
 
     if (!START_OF_TURN_SPELLS.includes(spellSlug)) return;
 
-    const spell = fromUuidSync(spellUuid) as SpellPF2e;
+    let spell = fromUuidSync(spellUuid) as SpellPF2e;
+
+    // for spells cast from item activations (pf2e-dailies), get the spell object from the message
+    if (!spell || spell.slug !== spellSlug) {
```
olilan1: What is this check `spell.slug !== spellSlug` for? Is it possible for the spell to be retrieved, but for its slug to not match?

When I tried this, it seems like the problem is the uuid isn't right, so spell will just be null here I thought.

src/startofturnspells.ts
```diff
@@ -23,12 +23,21 @@ export async function addEffectsToTokensInStartOfTurnTemplates(template: Measure
...
     const spellSlug = template.flags.pf2e?.origin?.slug;
-
     const spellUuid = template.flags.pf2e?.origin?.uuid as string;
 
     if (!START_OF_TURN_SPELLS.includes(spellSlug)) return;
 
-    const spell = fromUuidSync(spellUuid) as SpellPF2e;
+    let spell = fromUuidSync(spellUuid) as SpellPF2e;
+
+    // for spells cast from item activations (pf2e-dailies), get the spell object from the message
```
olilan1: This isn't just for when pf2e-dailies is in use. It also happens with a wand of a spell, which gets automatically added to the Activations tab by the PF2e system I think.

src/startofturnspells.ts
```diff
@@ -133,6 +142,9 @@ export async function postMessagesForWithinEffects(combatant: CombatantPF2e) {
         const spellUuid = effect.flags["samioli-module"]?.startOfTurnSpellUuid as string;
         const spell = fromUuidSync(spellUuid) as SpellPF2e;
         const content = `${combatant.name} has started their turn within ${spell?.name}`
+        if (combatant.tokenId) {
+            await replaceTargets([combatant.tokenId]);
+        }
```
olilan1, line 163: What was this change needed for? Is it related to the Activations-tab spells?

mustangsami: No, it was because I realised it didn't target the token that was targeted by the spell. So this way it adds it to the chat message when it's created.

src/startofturnspells.ts
mustangsami: ok, so just remove "pf2e-dailies" reference?

src/startofturnspells.ts
```diff
@@ -107,14 +113,22 @@ export async function addOrRemoveWithinEffectIfNeeded(token: TokenPF2e, costInFe
...
+                const spellSource = template.document.getFlag(
+                    "samioli-module",
+                    "spellSource"
+                ) as SpellSource | undefined;
+                const casterUuid = template.document.getFlag(
+                    "samioli-module",
+                    "casterUuid"
+                ) as string | undefined;
+                const spell = getSpellOrFallback(spellUuid, spellSource, casterUuid);
                 if (!spell) return;
```
olilan1: This line was here before, but there's an issue - we are inside a `for` loop here, and if we return we will exit the whole function and the rest of the loop won't execute.

Should be `continue` instead to carry on with the rest of the loop.

src/startofturnspells.ts
```diff
@@ -161,12 +177,45 @@ export async function deleteWithinEffectsForTemplate(template: MeasuredTemplateD
...
+    if (spellSource) {
+        const caster = casterUuid ? fromUuidSync(casterUuid) as ActorPF2e : null;
+        const spellViaFallback = new CONFIG.Item.documentClass(
+            spellSource,
+            { parent: caster }
+        ) as SpellPF2e;
+
+        if (caster) {
+            // Inject synthetic spell in-memory for standard rolls.
+            (caster.items as unknown as Map<string, ItemPF2e>).set(
```
olilan1, line 214: Do we need to change the actual caster's items like this? I'd have thought that might cause issues if something tries to interact with the caster and finds this extra item there. (E.g. does this affect the character sheet?)

Does it not work with just the spell object created above?

src/startofturnspells.ts
```diff
@@ -1,46 +1,52 @@
...
     }
 
     if (!spell) return;
 
-    await template.setFlag("samioli-module", 'isStartOfTurnSpell', true);
+    await template.setFlag("samioli-module", "isStartOfTurnSpell", true);
+    if (isTransient) {
+        await template.setFlag("samioli-module", "spellSource", spell.toObject());
+        await template.setFlag("samioli-module", "casterUuid", spell.actor?.uuid);
+    }
```
olilan1: This is potentially three separate database updates that we have to `await` for. Could we do this instead?

``` TS
    const flags: Record<string, unknown> = {
        isStartOfTurnSpell: true,
    };
    if (isTransient) {
        flags.spellSource = spell.toObject();
        flags.casterUuid = spell.actor?.uuid;
    }
    await template.update({ "flags.samioli-module": flags });
```

src/startofturnspells.ts
mustangsami: It doesn't effect the character sheet as far as I can tell. I've looked pretty much everywhere. I can see it on the actor in debug but nothing on the character sheet. I've added in something to clean up the item after the template is removed just in case also.

It doesn't work if it's not added, as the logic for the damage roll requires there to be an actual "item" on the actor that it can reference otherwise it doesn't work when attempting to roll damage from a recreated spell that's been put into chat.

In other words:

1. Spell cast from activations
2. Template placed
3. Creature is given the effect that it's in a start of turn spell
4. Turn progressed to the creature
5. Chat message of the spell created
6. Roll for damage (this is where it fails without injecting the item)

olilan1: ok, I see. Having the cleanup step seems good.

olilan1 (approved):

PR #136, opened 2026-05-24
Automated Mirror image

mustangsami:
Full automation for Mirror Image included animation.

Note: removal of existing AA macro is needed because they will compete.

### DIFF
--- a/src/spells/mirrorimage.ts
+++ b/src/spells/mirrorimage.ts
@@ -148,23 +148,23 @@ export async function handleMirrorImageCreated(item: EffectPF2e): Promise<void> 
 
     const seq = new Sequence()
         .effect()
-        .file("jb2a.impact.004.blue")
-        .atLocation(token)
-        .fadeIn(500)
-        .tieToDocuments([item])
-        .randomRotation()
-        .fadeOut(1500)
+            .file("jb2a.impact.004.blue")
+            .atLocation(token)
+            .fadeIn(500)
+            .tieToDocuments([item])
+            .randomRotation()
+            .fadeOut(1500)
         .effect()
-        .file("jb2a.extras.tmfx.runes.circle.simple.illusion")
-        .atLocation(token)
-        .duration(2000)
-        .fadeIn(500)
-        .fadeOut(1500)
-        .tieToDocuments([item])
-        .scale(0.5)
-        .filter("Glow", { color: 0x0096ff })
-        .scaleIn(0, 500, { ease: "easeOutCubic" })
-        .waitUntilFinished(-1000);
+            .file("jb2a.extras.tmfx.runes.circle.simple.illusion")
+            .atLocation(token)
+            .duration(2000)
+            .fadeIn(500)
+            .fadeOut(1500)
+            .tieToDocuments([item])
+            .scale(0.5)
+            .filter("Glow", { color: 0x0096ff })
+            .scaleIn(0, 500, { ease: "easeOutCubic" })
+            .waitUntilFinished(-1000);
 
     for (let i = 0; i < imagesCount; i++) {
         seq.addSequence(getMirrorImageSequence(i, item, token));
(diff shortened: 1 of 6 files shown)


src/spells/mirrorimage.ts
```diff
@@ -0,0 +1,243 @@
...
+    for (let i = 0; i < imagesCount; i++) {
+        seq.addSequence(getMirrorImageSequence(i, item, token));
+    }
+
+    await seq.play();
+}
+
+export async function handleMirrorImageUpdated(
+    item: EffectPF2e,
+    changes: MirrorImageChanges
```
olilan1: As this is called from the hook whenever the effect changes, it's possible the effect could change in ways that aren't covered by this `MirrorImageChanges` interface (which just includes the badge value).

I think it'd be better to use `DeepPartial<EffectSource>`  here, which would allow you to remove the `MirrorImageChanges` interface.

src/spells/mirrorimage.ts
```diff
@@ -0,0 +1,243 @@
...
+
+/** Decrements the mirror image badge count, or deletes the effect if it reaches 0. */
+async function decrementOrDeleteEffect(effect: EffectPF2e): Promise<void> {
+    const badgeVal = Number(effect.system.badge?.value ?? 0);
+    if (badgeVal > 1) {
+        await effect.update({ "system.badge.value": badgeVal - 1 });
+    } else {
+        await effect.delete();
+    }
+}
```
olilan1: I don't think this is necessary - you can just do `effect.decrease()` and it will decrement it, and remove the effect if it hits zero.

src/spells/mirrorimage.ts
```diff
@@ -0,0 +1,243 @@
...
+
+export async function checkForMirrorImage(chatMessage: ChatMessagePF2e): Promise<void> {
```
olilan1: As there's more than one roll defined for this spell, it might be good to put the word "Attack" in the name. Maybe `checkForMirrorImageOnAttack`?

src/spells/mirrorimage.ts
```diff
@@ -0,0 +1,243 @@
...
+
+    return new Sequence()
+        .effect()
+            .name(`Mirror Image Nr.${1 + number}`)
+            .copySprite(token)
+            .origin(item.uuid)
+            .fadeIn(1000)
+            .tieToDocuments([item])
+            .fadeOut(1000)
+            .attachTo(token, { followRotation })
```
olilan1: I don't think this is wrong, but I'm seeing a TS error on this line (because the Sequencer types aren't complete?). If you're seeing this as well, shall we add a `ts-ignore-error` here?

mustangsami: Not seeing the error on my mine? Do I need to update something? Or maybe you need to update something?

olilan1: I still see it on mine (after doing `npm install`). From what I can see, `AttachToOptions` in Sequencer's types doesn't include something called `followRotation` (though as I said, I think it is correct).

olilan1: Hmm, apparently this is happening because it's now called `bindRotation`.

And the default is true, which may be why this works anyway. This `followRotation` here doesn't seem to do anything.

Also I don't think we need an equivalent of `followRotation = !token.document.lockRotation` - shouldn't the animation always follow the token's rotation?

In summary, I think we can just remove this parameter.

src/spells/mirrorimage.ts
```diff
@@ -0,0 +1,225 @@
...
+    if (isNaN(badgeValue)) return;
+
+    const token = item.actor?.getActiveTokens()[0];
+    if (!token) return;
+
+    const activeEffects = Sequencer.EffectManager.getEffects({ origin: item.uuid });
+    const currentOnScreen = activeEffects.length;
+
+    if (badgeValue < currentOnScreen) {
+        const effectToEnd = activeEffects[activeEffects.length - 1];
```
olilan1: Unlike in the other case, here you only remove one animation when the badge value decreases. If it (somehow) decreases by more than one, you still only remove one animation.

olilan1 (approved):

PR #147, opened 2026-06-11
Courageous anthem

mustangsami:


src/utils.ts
```diff
@@ -401,3 +410,122 @@ export function getTokenFromUuid(uuid: string | null): TokenPF2e | null {
...
+    // Check alliance if set
+    if (targetActor.alliance && casterActor.alliance) {
+        return targetActor.alliance === casterActor.alliance;
+    }
+
+    // Fall back to disposition
+    const disposition = casterToken?.document.disposition ??
+        casterActor.getActiveTokens()[0]?.document.disposition;
+    return targetToken.document.disposition === disposition;
+}
```
olilan1: I don't think this function should be needed. From what I can tell, the disposition field isn't used in PF2e, so you're meant to just check alliances, and you can do that with a built-in method `actor.isAllyOf(otherActor)`

src/utils.ts
```diff
@@ -401,3 +410,122 @@ export function getTokenFromUuid(uuid: string | null): TokenPF2e | null {
...
+): boolean {
+    const deadCondition = options.deadCondition ?? "dead";
+    const unconsciousCondition = options.unconsciousCondition ?? "unconscious";
+    const minHp = options.minHp ?? 0;
+
+    const hp = actor.system.attributes.hp?.value ?? 0;
+    const isDead = actor.conditions.has(deadCondition) || hp <= minHp;
+    const isUnconscious = actor.conditions.has(unconsciousCondition);
+    return !isDead && !isUnconscious;
+}
```
olilan1: I don't think this function works, I think `actor.conditions.has("unconscious")` isn't the right way to check for this.

I think you can use PF2e helper functions to do this. `actor.hasCondition("unconscious")` works instead. And for dead, there is specifically `actor.isDead`.

Also I think this could be much simpler - why have this `options` parameter that you're not using? Why would you want to povide a different string for the condition that is "dead", when that condition is always called "dead"?

 This whole function could be just:
``` TS
return !actor.isDead && !actor.hasCondition("unconscious");
```

(I'm not sure if you need the HP > 0 check because from what I can see, when an actor's health is 0, isDead will return true or they will be unconscious, though there might be some case I'm missing)

src/utils.ts
```diff
@@ -401,3 +410,122 @@ export function getTokenFromUuid(uuid: string | null): TokenPF2e | null {
...
+    const backend = CONFIG.Canvas.polygonBackends[collisionType];
+    if (!backend) return false;
+
+    const hasCollision = backend.testCollision(
+        originPoint,
+        targetPoint,
+        { type, mode }
+    );
+    return !hasCollision;
+}
```
olilan1: For Courageous Anthem, this seems too strict to me, though I guess it could come down to how you interpret the rules.

This seems to check if there's an unbroken line from the center of the caster's square to the center of the target's square. But that means if they are diagonally next to each other but there's a single wall piece, it won't work. It also won't work if someone is just around a corner.

That might be right for a fireball, but as courageous anthem is sound-based, I think most GMs would say it should work if someone is just around a corner.

See image for a case that currently doesn't work.

Is it even worth checking line of effect for this?

Image

src/utils.ts
```diff
@@ -401,3 +410,122 @@ export function getTokenFromUuid(uuid: string | null): TokenPF2e | null {
...
+/**
+ * Checks if the distance between two tokens or points is within a specified number of feet.
+ */
+export function isWithinDistance(
+    origin: TokenPF2e | Point,
+    target: TokenPF2e | Point,
+    maxDistanceFeet: number
+): boolean {
+    return getDistance(origin, target) <= maxDistanceFeet;
+}
```
olilan1: Again I don't think this is needed. Instead of your `getDistance` function, you can just do:

`token.distanceTo(otherToken)`

And this `isWithinDistance` isn't really needed, as the whole check in the courageous anthem code can be just:
``` TS
if (token.distanceTo(t) > 60) return false;
```

src/spells/courageousanthem.ts
```diff
@@ -0,0 +1,139 @@
...
+        return;
+    }
+
+    // Find all allies within 60 feet on the current scene who have line of effect and are alive
+    const allTokens = canvas.tokens.placeables as TokenPF2e[];
+    const targetTokens = allTokens.filter(t => {
+        if (t === token) return true; // Caster is always included
+
+        const actor = t.actor;
+        if (!actor || (actor.type !== "character" && actor.type !== "npc")) {
```
olilan1: I think this excludes familiars.

Also you can use the helper function `actor.isOfType` to make this easier, and there's a synthetic type "creature" that is true for characters, npcs, and familiars.

So this can be just:

``` TS
if (!actor?.isOfType("creature")) return false;
```

src/utils.ts
```diff
@@ -203,17 +203,26 @@ export async function createTemplateAtPoint(point: Point, userId: string, radius
...
+            return existingEffect;
+        }
     }
-    await actor.createEmbeddedDocuments('Item', [effectSource]);
-    const newEffect = actor.items.find(item => item.slug === effectSource.system.slug
-        && item.type === 'effect');
+    await actor.createEmbeddedDocuments("Item", [sourceClone]);
+    const newEffect = actor.items.find(
+        item => item.slug === sourceClone.system.slug && item.type === "effect"
+    );
```
olilan1: You shouldn't need to do a `find` to get the thing that was just created - `createEmbeddedDocuments` returns what it creates.

You can do this instead:

``` TS
const [newEffect] = await actor.createEmbeddedDocuments("Item", [sourceClone]) as EffectPF2e[];
```

src/utils.ts
```diff
@@ -401,3 +407,19 @@ export function getTokenFromUuid(uuid: string | null): TokenPF2e | null {
...
+
+/**
+ * Checks if a target token is an ally of the caster.
+ * Compares alliance values, falling back to disposition if not set.
+ * 
+ * Note on disposition fallback:
+ * In Foundry, token disposition indicates allegiance (Friendly = 1, Neutral = 0, Hostile = -1).
+ * If actor alliance information is missing or unconfigured, tokens with matching disposition
+ * values (e.g. friendly caster targeting a friendly ally) are considered allies.
+ */
```
olilan1: Rogue left-over comment here

olilan1 (approved):

PR #148, opened 2026-06-12
Removed sustain logic from spells with the summon trait

mustangsami:


src/sustain.ts
```diff
@@ -138,16 +139,34 @@ function getTemplateFromEffect(effect: EffectPF2e) {
     return canvas.scene?.templates.get(templateId);
 }
 
-async function addSustainedSpellBackIntoChat(effect: EffectPF2e, actor: ActorPF2e) {
+async function addSustainedSpellBackIntoChat(effect: EffectPF2e) {
```
olilan1: As this function now doesn't necessarily add the spell back into the chat, perhaps it should be called something like `postSustainChatMessage` instead

src/sustain.ts
```diff
@@ -138,16 +139,34 @@ function getTemplateFromEffect(effect: EffectPF2e) {
...
+    if (!spell) return;
+
+    // Do not add spell message back into chat if spell is a summon
+    // Instead, we add a message to confirm that it's been sustained
+    if (spell.traits.has('summon')) {
+        const content = `<p><strong>${spell.name}</strong> was sustained.</p>`;
+        await ChatMessage.create({
+            content: content,
+            speaker: ChatMessage.getSpeaker({ actor: spell.actor })
+        });
```
olilan1: Can use our util function `sendBasicChatMessage` for this.

src/sustain.ts
```diff
@@ -138,16 +139,34 @@ function getTemplateFromEffect(effect: EffectPF2e) {
...
+async function addSustainedSpellBackIntoChat(effect: EffectPF2e) {
     const skipChat = !!effect.getFlag(MODULE_ID, "skipSustainChat");
     if (skipChat) return;
 
+    const spell = getSpellFromEffect(effect)
+    if (!spell) return;
+
+    // Do not add spell message back into chat if spell is a summon
+    // Instead, we add a message to confirm that it's been sustained
+    if (spell.traits.has('summon')) {
```
olilan1: It might be good to do this in the case where `skipChat` (a few lines above) is true as well. The `skipSustainChat` thing is currently used only for Dancing Blade I think, but having this basic chat message seems like a good idea for any case where we don't want to post the spell again.

src/sustain.ts
```diff
@@ -167,19 +186,68 @@ async function createSustainChatMessage(actor: ActorPF2e, spell: SpellPF2e, effe
...
 }
 
+export async function handleRemoveSummon(casterId: string) {
+    if (typeof casterId !== "string" || !casterId) return;
+
+    const tokens = canvas.tokens?.placeables ?? [];
+    const summons = tokens.filter(token => {
+        const actor = token.actor;
+        if (!actor) return false;
+        const summonerId = actor.getFlag("pf2e-summons-assistant", "summoner.id");
```
olilan1, line 223: Does this only work if you're also using pf2e-summons-assistant? I don't think want to have a dependency on that module. We can have functionality that works with it, but it should behave nicely if that module is not installed.

At the moment, it looks like if you don't have pf2e-summons-assistant, you'll see the button to remove the summon, but when you press it you'll just see a notification saying "No matching summoned tokens".

Perhaps we should only show the remove button if pf2e-summons-assistant is installed (or possibly, only if there is a valid token to remove, so we don't just show a button that doesn't work). If that module isn't installed, I'm not sure we should handle removing the summoned token when we're not handling adding the summoned token.

src/sustain.ts
```diff
@@ -167,19 +186,68 @@ async function createSustainChatMessage(actor: ActorPF2e, spell: SpellPF2e, effe
...
+    if (summons.length === 0) {
+        ui.notifications?.info("No matching summoned tokens found on the canvas.");
+        return;
+    }
+
+    for (const summon of summons) {
+        if (summon.id) {
+            getSocket().executeAsGM(DELETE_SUMMON, summon.id);
+        }
+    }
```
olilan1: So, if the caster has more than one summon on the board, we remove all of them? That doesn't seem right.

We need to know which summon relates to the spell that was not sustained. Not sure how we can do that, when we're not adding the summon token to the board ourselves.

I wonder if it's necessary to have this removal functionality, perhaps just the message "[spell] was not sustained" is enough, and the GM can remove the token.

mustangsami: I've added a dialog to select which summon should be removed. I think this is the best of both worlds.

src/sustain.ts
```diff
@@ -167,19 +181,113 @@ async function createSustainChatMessage(actor: ActorPF2e, spell: SpellPF2e, effe
...
 
     const spellName = item.name.replace('Sustaining: ', '');
     const content = `<p><strong>${spellName}</strong> was not sustained.</p>`;
+
+    const spell = getSpellFromEffect(item as EffectPF2e)
+    if (!spell) return;
+
+    const isSummonAssistantEnabled = game.modules.get('pf2e-summons-assistant')?.active;
+    const isSpellASummon = spell.traits.has('summon');
+    const casterHasActiveSummons = getSummonedTokensFromCanvas(item.actor?.id!)
```
olilan1: nit: `getSummonedTokensFromCanvas` returns a list of tokens, not a boolean.

As `casterHasActiveSummons` sounds like it should be a boolean, add `!!` to the front of this to turn it into a boolean.

src/sustain.ts
```diff
@@ -167,19 +181,113 @@ async function createSustainChatMessage(actor: ActorPF2e, spell: SpellPF2e, effe
...
+    const optionsHtml = summons
+        .map(s => new Option(s.name, s.id).outerHTML)
+        .join("");
+
+    const summonId = await DialogV2.wait({
+        window: { title: "Which Summon was not sustained?" },
+        content: `
+            <form>
+                <div class="form-group">
+                    <label>Select Summon:</label>
```
olilan1: When I tried this it split this text onto two lines which didn't look great. Could consider using a non-breaking space between the two words so they have to stay together. Also possibly make the dialog a bit wider?

Alternatively,  as the dialog title already asks the question, we could omit the label and just have the drop-down box in the main bit of the dialog. Up to you.

Image

src/sustain.ts
```diff
@@ -167,19 +181,113 @@ async function createSustainChatMessage(actor: ActorPF2e, spell: SpellPF2e, effe
...
+                    <select name="summonSelect">
+                        ${optionsHtml}
+                    </select>
+                </div>
+            </form>
+        `,
+        buttons: [{
+            action: "remove",
+            label: "Banish",
```
olilan1: I feel like this label should match the verb from the button in chat, so either both should say Remove, or both should say Banish.

olilan1 (approved):

PR #149, opened 2026-06-12
Automated Off guard application for Tumble Behind

mustangsami:


src/actions/tumblethrough.ts
```diff
@@ -184,95 +203,169 @@ async function animateSuccessfulTumble(token: TokenPF2e, target: TokenPF2e, rota
...
+        .waitUntilFinished(-250)
         .animation()
-            .on(token)
-            .opacity(1)
-            .fadeIn(200)
-            .duration(250);
+        .on(token)
+        .opacity(1)
+        .fadeIn(200)
+        .duration(250);
```
olilan1: Your formatting got messed up

src/actions/tumblethrough.ts
```diff
@@ -184,95 +203,169 @@ async function animateSuccessfulTumble(token: TokenPF2e, target: TokenPF2e, rota
...
+            rules: [
+                {
+                    key: "RollOption",
+                    domain: "all",
+                    option: "target:condition:off-guard",
+                    predicate: [`target:effect:tumble-behind-off-guard-${sluggifiedId}`]
+                }
+            ]
+        }
+    } as unknown as EffectSource;
```
olilan1: Am I right in thinking that these effects don't make sure that the benefit only applies to the next attack you make? So if you Tumble Behind, then attack twice, it would be off guard against both rather than just the first?

I think this can be fixed, but it requires some changes. The effect can be removed after the roll, but only on the attacker's effect. So, we need the bonus to be applied through an effect on the attacker.

Fortunately, this is possible. I saw this by looking at the existing effect for Tumble Behind provided by the system. It uses an `EphemeralEffect` rule, to apply the actual Off Guard effect to the target. We can do the same thing.

There's another issue which is that `removeAfterRoll` doesn't seem to work on `EphemeralEffect`. So we have to add a dummy `FlatModifier` effect to make it happen.

We can also use the trick you used with the actor signature to make the effect apply only when attacking the specific target, without needing to apply an effect to them.

So, all in all, this should do it, with no effect on the target at all:

``` TS
    const targetSignature = targetToken.actor.signature;

    const attackerEffectSource = {
        type: "effect",
        name: "Tumble Behind (Attacker)",
        img: " -- SOME ICON, PROBABLY NOT THE OFF-GUARD ONE THOUGH -- ",
        system: {
            slug: "tumble-behind-attacker",
            tokenIcon: { show: false },
            duration: {
                value: 0,
                unit: "rounds",
                expiry: "turn-end"
            },
            rules: [
                {
                    key: "EphemeralEffect",
                    selectors: ["attack-roll"],
                    uuid: "Compendium.pf2e.conditionitems.Item.AJh5ex99aV6VTggg",
                    predicate: [`target:signature:${targetSignature}`]
                },
                {
                    key: "FlatModifier",
                    selector: "attack-roll",
                    value: 0,
                    type: "untyped",
                    predicate: [`target:signature:${targetSignature}`],
                    removeAfterRoll: "if-enabled"
                }
            ]
        }
    } as unknown as EffectSource;
```

mustangsami: My concern around this was for hero point rerolls. If it's automatically removed, it would need to be reapplied. We had the same issue with panache. I could maybe instead add a button to remove it?

src/actions/tumblethrough.ts
```diff
@@ -169,7 +180,7 @@ async function animateFailureTumble(x: number, y: number, token: TokenPF2e, targ
         .sound()
             .file(landSound)
             .delay(knockedBackAnimTime - 500)
-        .animation()
+            .animation()
```
olilan1: formatting still not right here

src/actions/tumblethrough.ts
```diff
@@ -187,7 +198,7 @@ async function animateSuccessfulTumble(token: TokenPF2e, target: TokenPF2e, rota
             .copySprite(token)
             .duration(500)
             .waitUntilFinished(-450)
-        .animation()
+            .animation()
```
olilan1: here also

olilan1 (approved):

PR #154, opened 2026-06-28
Refactor hook runner to allow functions to be called only when needed

olilan1:
This changes the hook calls in `hooks.ts` to include checks so that the functions are only called when needed.

This makes it easier to see when code is being executed, and ensures that we are not unnecessarily running code on every hook, e.g.  on every chat message.

All HookRunner calls must now include these guard checks, unless called with `.allowUnfilteredRun()`.

### DIFF
--- a/src/effects/panache.ts
+++ b/src/effects/panache.ts
@@ -159,7 +159,9 @@ export function isParryOrBuckleEligible(chatMessage: ChatMessagePF2e): boolean {
   const target = chatMessage.target?.actor;
   if (!context || !target) return false;
 
-  const hasDuelingParry = context.options.includes("target:effect:dueling-parry");
+  const hasDuelingParry =
+    context.options.includes("target:effect:dueling-parry") ||
+    context.options.includes("target:effect:extravagant-parry");
   const hasShieldRaised = context.options.includes("target:effect:raise-a-shield");
   const isFailure = context.outcome === "failure";
   const isCriticalFailure = context.outcome === "criticalFailure";
--- a/src/effects/panache.ts
+++ b/src/effects/panache.ts
@@ -154,22 +154,22 @@ function getPanacheItems(actor: ActorPF2e) {
   return items.filter(item => item.type === "effect" && item.system.slug === "effect-panache");
 }
 
-export function isParryOrBuckleEligible(chatMessage: ChatMessagePF2e): boolean {
+export function isPanacheGeneratingParryOrBuckler(chatMessage: ChatMessagePF2e): boolean {
   const context = chatMessage.flags.pf2e.context;
   const target = chatMessage.target?.actor;
   if (!context || !target) return false;
 
-  const hasDuelingParry =
-    context.options.includes("target:effect:dueling-parry") ||
-    context.options.includes("target:effect:extravagant-parry");
-  const hasShieldRaised = context.options.includes("target:effect:raise-a-shield");
+  const hasExtravagantParry =
+    (context.options?.includes("target:effect:dueling-parry") ||
+    context.options?.includes("target:effect:extravagant-parry")) ?? false;
+  const hasShieldRaised = context.options?.includes("target:effect:raise-a-shield") ?? false;
   const isFailure = context.outcome === "failure";
   const isCriticalFailure = context.outcome === "criticalFailure";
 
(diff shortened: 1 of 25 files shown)


src/hooks.ts
```diff
@@ -1,38 +1,87 @@
...
 import { startTumbleThrough } from "./actions/tumblethrough.ts";
-import { editEnjoyTheShowSkillRollIfNeeded, startEnjoyTheShow } from "./actions/enjoytheshow.ts";
-import { checkForBravado, checkForExtravagantParryOrElegantBuckler, checkForFinisherAttack, checkForFinisherDamage } from "./effects/panache.ts";
+import { editEnjoyTheShowSkillRoll, startEnjoyTheShow } from "./actions/enjoytheshow.ts";
+import {
+    applyPanacheForActor,
+    applyPanacheForParryOrBuckler,
+    handleFinisherAttack,
+    clearPanacheForActor,
+    isParryOrBuckleEligible
```
mustangsami: nit: typo

src/effects/panache.ts
```diff
@@ -173,35 +154,30 @@ function getPanacheItems(actor: ActorPF2e) {
   return items.filter(item => item.type === "effect" && item.system.slug === "effect-panache");
 }
 
-export async function checkForExtravagantParryOrElegantBuckler(chatMessage: ChatMessagePF2e) {
+export function isParryOrBuckleEligible(chatMessage: ChatMessagePF2e): boolean {
```
mustangsami: nit: typo

src/hookrunner.ts
mustangsami, line 1: Might be nice to put some comments on the helper functions to explain the usage

src/hooks.ts
```diff
@@ -42,69 +91,103 @@ Hooks.on("init", () => {
...
             .ifEnabled(SETTINGS.TEMPLATE_TARGET)
+            .allowUnfilteredRun()
             .run();
     }
 
-    hook(checkIfTemplatePlacedHasSustainEffect, template)
+    hook(associateTemplateWithSustainedEffect, template)
         .ifEnabled(SETTINGS.AUTO_SUSTAIN_CHECK)
         .ifGM()
+        .if(() => template.actor?.items.some(
```
mustangsami: Might be nicer to have this be an exported function within sustain

mustangsami: I'm seeing a few of these around actually. There are some situations where the `.if(() =>` is calling a function and others where it just resolves it inline.

I'm wondering if it would be nicer to just always have it be a nicely named exported function that explains clearly what it does so it's easy to understand.

mustangsami (approved):

src/hooks.ts
olilan1: Renamed to `isPanacheGeneratingParryOrBuckler`

src/effects/panache.ts
olilan1: Renamed to `isPanacheGeneratingParryOrBuckler`

src/hooks.ts
olilan1: As mentioned, I left it that way for some of the template ones, but I think you're right that defining functions for them anyway makes it easier to understand. Done.

src/hooks.ts
```diff
@@ -1,38 +1,90 @@
...
 import { handleFrightenedAtTurnEnd } from "./effects/frightened.ts";
-import { addButtonClickHandlersIfNeeded } from "./chatbuttonhelper.ts";
-import { postMessagesForWithinEffects, deleteWithinEffectsForTemplate, addEffectsToTokensInStartOfTurnTemplates, addOrRemoveWithinEffectIfNeeded } from "./startofturnspells.ts";
+import { addButtonClickHandlers } from "./chatbuttonhelper.ts";
+import {
+    postMessagesForWithinEffects,
+    deleteWithinEffectsForTemplate,
+    addEffectsToTokensInStartOfTurnTemplates,
+    addOrRemoveWithinEffectIfNeeded,
+    START_OF_TURN_SPELLS,
```
mustangsami: I'm getting a `START_OF_TURN_SPELLS' is declared but its value is never read.` on this line

olilan1: oh yeah, me too

olilan1: Fixed

PR #158, opened 2026-06-28
Sustain bug fix when initiative is shared

mustangsami:
Fix for: https://github.com/olilan1/samioli-module/issues/153

### DIFF
--- a/src/sustain.ts
+++ b/src/sustain.ts
@@ -168,7 +168,6 @@ function getSpellFromEffect(effect: EffectPF2e): SpellPF2e | undefined {
     const spell = fromUuidSync(spellUuid);
     if (!(spell instanceof CONFIG.PF2E.Item.documentClasses.spell)) return;
     return spell;
-
 }
 
 async function createSustainChatMessage(actor: ActorPF2e, spell: SpellPF2e, effect: EffectPF2e) {
@@ -204,7 +203,6 @@ export async function createSpellNotSustainedChatMessage(item: ItemPF2e) {
     // if spell is a summon check if there are relevant summons
     // create a chat with a button to remove the summoned token
     if (isSpellASummon && casterHasActiveSummons) {
-        if (!item.actor) return;
         await createChatMessageWithButton({
             slug: "remove-summon",
             actor: item.actor,
--- a/src/sustain.ts
+++ b/src/sustain.ts
@@ -200,13 +200,12 @@ export async function createSpellNotSustainedChatMessage(item: ItemPF2e) {
     const spell = getSpellFromEffect(item as EffectPF2e)
     if (!spell) return;
 
-    const isSummonAssistantEnabled = game.modules.get('pf2e-summons-assistant')?.active;
     const isSpellASummon = spell.traits.has('summon');
-    const casterHasActiveSummons = !!getSummonedTokensFromCanvas(item.actor?.id!)
+    const casterHasActiveSummons = !!getSummonedTokensFromCanvas(item.actor?.id!);
 
-    // if spell is a summon, module is active and there are relevant summons on the canvas 
-    // create a chat with a button to remove the summoned token    
-    if (isSummonAssistantEnabled && isSpellASummon && casterHasActiveSummons) {
+    // if spell is a summon check if there are relevant summons
+    // create a chat with a button to remove the summoned token
+    if (isSpellASummon && casterHasActiveSummons) {
         if (!item.actor) return;
         await createChatMessageWithButton({
             slug: "remove-summon",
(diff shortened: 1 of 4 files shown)


src/sustain.ts
```diff
@@ -192,13 +196,12 @@ export async function createSpellNotSustainedChatMessage(item: ItemPF2e) {
     const spell = getSpellFromEffect(item as EffectPF2e)
     if (!spell) return;
 
-    const isSummonAssistantEnabled = game.modules.get('pf2e-summons-assistant')?.active;
     const isSpellASummon = spell.traits.has('summon');
-    const casterHasActiveSummons = !!getSummonedTokensFromCanvas(item.actor?.id!)
+    const casterHasActiveSummons = !!getSummonedTokensFromCanvas(item.actor?.id!);
```
olilan1: This check doesn't work as intended when the array is empty.

An empty array is "truthy" in Javascript, i.e. this will resolve to `true` when `getSummonedTokensFromCanvas` returns an empty array - which it now does whenever the summons module isn't active.

So, the "Remove summons" button will appear all the time if the spell is a summon spell.

Instead, check `.length > 0`

src/sustain.ts
```diff
@@ -337,3 +342,25 @@ export function hasSustainingEffect(template: MeasuredTemplateDocumentPF2e): boo
         i => i.type === "effect" && (i.slug?.startsWith("sustaining-effect-") ?? false)
     ) ?? false;
 }
+
+async function resetSustainFlagsForActor(actor: ActorPF2e) {
+    const sustainedEffects = getActorSustainedEffects(actor);
+    if (!sustainedEffects) return;
+
+    for (const effect of sustainedEffects) {
```
olilan1: We already have a loop over the sustained effects above, so by doing this we are calling `getActorSustainedEffects` twice and looping over it again.

Instead, we could make this function take an effect and just check and update the flag, and do this inside the loop that already exists at line 92.

olilan1 (commented):
The main change to sustain now makes it work fine in the case where the following actor has the same initiative, but not work in all other cases, because the removal now occurs twice.

I guess the expiration / auto-removal and the new removal are both occurring, because things are happening asynchronously.

One thing that occurs to me is that with the new approach we no longer need to set the duration of these sustained spells to 1 round. It might be better to set them to the actual (maximum) duration, e.g. for a spell that is "sustained up to 1 minute", can we set the duration to 1 minute?

There may still be an issue though when the duration expires - we need to make sure we don't double-remove the effect when it expires. One possible approach might be to check for unsustained spells in the `updateCombat` hook instead, which might happen after the effect expiry is checked, but I haven't verified that.

src/sustain.ts
```diff
@@ -181,28 +183,27 @@ async function createSustainChatMessage(actor: ActorPF2e, spell: SpellPF2e, effe
...
-    const casterHasActiveSummons = !!getSummonedTokensFromCanvas(item.actor?.id!)
+    const casterHasActiveSummons = getSummonedTokensFromCanvas(item.actor.id).length > 0;
 
-    // if spell is a summon, module is active and there are relevant summons on the canvas 
-    // create a chat with a button to remove the summoned token    
-    if (isSummonAssistantEnabled && isSpellASummon && casterHasActiveSummons) {
+    // if spell is a summon check if there are relevant summons
+    // create a chat with a button to remove the summoned token
+    if (isSpellASummon && casterHasActiveSummons) {
         if (!item.actor) return;
```
olilan1: I think this line can now be removed as you've added an `if (!item.actor) return;` check above

src/sustain.ts
```diff
@@ -163,10 +164,11 @@ async function postSustainChatMessage(effect: EffectPF2e) {
 function getSpellFromEffect(effect: EffectPF2e): SpellPF2e | undefined {
     const spellId = effect.getFlag(MODULE_ID, "sustainedSpellId");
     if (typeof spellId !== "string" || !spellId) return;
-    const spellUuid = 'Actor.' + effect.actor!.id + '.Item.' + spellId;
+    const spellUuid = `${effect.actor!.uuid}.Item.${spellId}`;
     const spell = fromUuidSync(spellUuid);
     if (!(spell instanceof CONFIG.PF2E.Item.documentClasses.spell)) return;
     return spell;
+
```
olilan1: rogue extra line here?

olilan1 (approved):

src/sustain.ts
olilan1: Removed it

src/sustain.ts
olilan1: Removed it

